PR #6, opened 2026-05-12
ADFA-2433 | random-xkcd: new demo plugin

fryanpan:
## Description

Adds the `random-xkcd` plugin as a small plugin example that demonstrates a few concepts for building plugins.

- Tutorial takes the user through key parts of how to make this plugin (starting from the "Plugin" template in CodeOnTheGo itself)
- UI integration as a bottom-sheet tab
- Fetches a random xkcd comic over HTTPS (requires network permission)
- Single-tap fetches a new one; double-tap copies the URL to the clipboard; triple-tap copies the image via the host's `FileProvider` authority
- Long-press the tab to surface a Tier-1 / Tier-2 / Tier-3 walkthrough served from `assets/docs/`

Can review as a whole PR, or follow the commits that go through the same steps as the tutorial doc.

Recording from Claude going through all flows using [mobile-mcp](https://github.com/mobile-next/mobile-mcp) (and I also tested manually)

https://github.com/user-attachments/assets/868b76c8-1f4e-4b08-b5c9-b69454a8ef88

## Ticket

[ADFA-2433](https://appdevforall.atlassian.net/browse/ADFA-2433)

## Observation

xkcd CC BY-NC 2.5 attribution is surfaced in the panel (always-visible credit line beneath every comic) and in the Tier-3 docs.

Had to make a corresponding minor change in CodeOnTheGo repo. See matching PR here: appdevforall/CodeOnTheGo#1297

### Some open questions for the future

- **Localized Strings.** Hardcoded the tab title to be `"XKCD"`. The current Plugin API / `TabItem` doesn't provide a way to localize from `Context` (all the other plugins also hardcode strings for now).
- **Tab Tooltip Wiring.** Had to make a minor change in the paired CodeOnTheGo PR to make the 3-tier tooltip work. See matching PR: https://github.com/appdevforall/CodeOnTheGo/pull/1297

## Build + Test Status

- `./gradlew testDebugUnitTest` — green (10/10 in `TapCountClassifierTest`)
- `./gradlew assemblePlugin` — green; `random-xkcd.cgp` builds at 5.0 MB (release)
- Claude smoke-tested using mobile-mcp on Samsung Galaxy A56 (CoGo from the matching PR above, Android 16). Clean install + happy path; recording attached above.
- Manually tested the same flow too.

[ADFA-2433]: https://appdevforall.atlassian.net/browse/ADFA-2433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


random-xkcd/src/main/kotlin/com/codeonthego/xkcdrandom/fragments/XkcdPanelFragment.kt
```diff
@@ -0,0 +1,353 @@
+package com.codeonthego.xkcdrandom.fragments
```
Daniel-ADFA: for consistency with the existing plugin examples, please use org.appdevforall.<plugin> as the package name

random-xkcd/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,62 @@
...
+            android:name="plugin.version"
+            android:value="${pluginVersion}" />
+
+        <meta-data
+            android:name="plugin.description"
+            android:value="A random xkcd comic in your bottom sheet. Tap for a new one, double-tap to copy URL, triple-tap to copy image. Demo plugin: see Tier 3 docs for a code walkthrough." />
+
+        <meta-data
+            android:name="plugin.author"
+            android:value="Code on the Go Team" />
```
Daniel-ADFA: Please use App Dev For All as plugin author

Daniel-ADFA:
I wonder if having the controls just like it is on the XKCD website is better UX, what do you think @fryanpan 


Screenshot 2026-05-13 at 08 24 49

fryanpan:
> I wonder if having the controls just like it is on the XKCD website is better UX, what do you think @fryanpan
> 
> Screenshot 2026-05-13 at 08 24 49

Yes, that would be nicer UX.  I can make that edit!

Wasn't happy that we had to add a custom tap handler either...makes the demo more complicated than it needs to be.  With buttons on screen for going to random, next, prev,etc. we only need to do single tap (copy url) and double tap (copy image) and I can get rid of the custom tap recognizer.  That will be nicer too.

fryanpan:
Updated with five buttons like XKCD, and simplified to one tap = copy url, two taps = copy image:
https://github.com/user-attachments/assets/9176202e-c682-4470-8b56-cb3b58096008

Daniel-ADFA (approved):

PR #20, opened 2026-06-03
ADFA-4035 | Fix sketch checkbox OCR parsing

jatezzz:
## Description

Improves the Sketch to UI plugin generation for checkbox groups by handling noisy OCR output more reliably.

This update fixes cases where checkbox labels were split into separate text nodes, OCR artifacts like `Joption` were generated, and noisy `textColor` annotations such as `textcalar: black` were not correctly converted into valid Android XML.

## Details

* Added checkbox widgets to region OCR processing.
* Merged nearby checkbox label fragments such as `Option` + `1` into `Option 1`.
* Prevented consumed label fragments from being rendered as extra `TextView` nodes.
* Normalized common checkbox label OCR artifacts, including `Joption` → `Option`.
* Added stricter parsing support for OCR variants of `textColor`.
* Added color validation for `android:textColor`, `android:background`, and `app:backgroundTint`.
* Prevented invalid values like `group1` from being emitted as Android color attributes.
* Added tests for checkbox label association, noisy `textColor` parsing, invalid color rejection, and checkbox group rendering.

https://github.com/user-attachments/assets/1b1446cd-ccd7-40c3-ac18-60ded523a898

## Ticket

[ADFA-4035](https://appdevforall.atlassian.net/browse/ADFA-4035)

## Observation

The fix was validated with the original sketch scenario. The generated XML now preserves `Option 1`, `Option 2`, and `Option 3`, keeps checked states, avoids extra `TextView` output, and correctly emits `android:textColor="#000000"` for the checkbox group.

[ADFA-4035]: https://appdevforall.atlassian.net/browse/ADFA-4035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sketch-to-ui-plugin/src/main/java/org/appdevforall/codeonthego/computervision/domain/TextAssociator.kt
+++ b/sketch-to-ui-plugin/src/main/java/org/appdevforall/codeonthego/computervision/domain/TextAssociator.kt
@@ -12,6 +12,9 @@ import kotlin.math.max
  */
 object TextAssociator {
     private const val OVERLAP_THRESHOLD = 0.6
+    private const val MIN_LINE_TOLERANCE = 12
+    private const val LABEL_FRAGMENT_MAX_GAP_MULTIPLIER = 2
+    private const val MIN_LABEL_FRAGMENT_MAX_GAP = 28
 
     fun assignTextToParents(parents: List<ScaledBox>, texts: List<ScaledBox>, allBoxes: List<ScaledBox>): List<ScaledBox> {
         val consumedTexts = mutableListOf<ScaledBox>()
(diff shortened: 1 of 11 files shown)


sketch-to-ui-plugin/src/main/java/org/appdevforall/codeonthego/computervision/domain/TextAssociator.kt
```diff
@@ -88,19 +80,69 @@ object TextAssociator {
...
     }
 
+    private fun collectLabelFragments(
+        widget: ScaledBox,
+        anchor: ScaledBox,
+        availableTexts: List<ScaledBox>,
+        consumedTexts: List<ScaledBox>
+    ): List<ScaledBox> {
+        val lineTolerance = max(anchor.h, widget.h).coerceAtLeast(12)
+        val maxGap = max(widget.h * 2, 28)
```
Daniel-ADFA: Please move magic numbers to constants

sketch-to-ui-plugin/src/main/java/org/appdevforall/codeonthego/computervision/domain/TextAssociator.kt
```diff
@@ -88,19 +80,69 @@ object TextAssociator {
...
+
+        val fragments = mutableListOf(anchor)
+        var previous = anchor
+
+        while (true) {
+            val next = availableTexts
+                .asSequence()
```
Daniel-ADFA: Can while (true) ever spin forever here. e.g. if minByOrNull keeps returning a candidate that doesn't advance previous? Worth guarding the exit / hoisting the invariant filters out.

jatezzz: Good point. The current loop should terminate because each selected fragment is added to `fragments` and excluded from subsequent iterations, and `availableTexts` is finite. However, I agree that `while (true)` makes the progression less explicit.

Daniel-ADFA (approved):

PR #25, opened 2026-06-08
ADFA-4267 | Fix noisy OCR XML generation

jatezzz:
## Description

This PR improves the `sketch-to-ui-plugin` XML generation when OCR annotations contain recognition errors or overlapping detections.

The change fixes cases where noisy OCR metadata caused incorrect Android XML, including duplicated password inputs, incorrect password attributes, wrong switch gravity, style metadata becoming visible text, and incorrect ImageView dimensions.

It also refactors the parser flow to make OCR sanitization, token mapping, value resolution, and widget-specific recovery easier to understand and maintain.

## Details

* Collapses overlapping `text_entry_box` detections so a single input does not generate multiple `EditText` widgets or an artificial horizontal `LinearLayout`.
* Recovers noisy password metadata into the correct `EditText` attributes.
* Preserves valid explicit `EditText` text while preventing leaked metadata from becoming visible text.
* Recovers `layoutgravity:centerhorizontal` as `center_horizontal`.
* Prevents `textColor: blue` from becoming a RadioButton label.
* Recovers compact image dimensions such as noisy `64dp` OCR values.
* Adds regression coverage for password parsing, overlapping detections, RadioGroup labels, Switch text/gravity, and ImageView dimensions.
* Refactors parser responsibilities into clearer sanitizer, token mapping, value resolving, and widget recovery steps.
* Names parser regexes and fuzzy thresholds to reduce magic values and improve readability.

### Before

https://github.com/user-attachments/assets/80ec4094-0c06-45bf-b3d7-a68cd3e226b4

### After

https://github.com/user-attachments/assets/fa20a918-96c9-4fe8-9129-f6436d38d0a7

## Ticket

[ADFA-4267](https://appdevforall.atlassian.net/browse/ADFA-4267)

## Observation

Focused regression tests were added for the corrected OCR/XML generation paths.

[ADFA-4267]: https://appdevforall.atlassian.net/browse/ADFA-4267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sketch-to-ui-plugin/src/main/java/org/appdevforall/codeonthego/computervision/domain/TextInputDetectionCollapser.kt
```diff
@@ -0,0 +1,78 @@
...
+import java.util.IdentityHashMap
+
+object TextInputDetectionCollapser {
+    private const val TEXT_ENTRY_LABEL = "text_entry_box"
+    private const val TEXT_ENTRY_CONTAINMENT_THRESHOLD = 0.85f
+    private const val TEXT_ENTRY_IOU_THRESHOLD = 0.50f
+
+    fun collapse(candidates: List<DetectionResult>): List<DetectionResult> {
+        val textInputs = candidates.filter { it.label == TEXT_ENTRY_LABEL }
+        if (textInputs.size <= 1) return candidates
```
Daniel-ADFA, line 14: This can be simplified to check if it's 1 item instead, if there's a need to check for < 1 then candidates will be an empty array in this case.

sketch-to-ui-plugin/src/main/java/org/appdevforall/codeonthego/computervision/domain/TextInputDetectionCollapser.kt
```diff
@@ -0,0 +1,78 @@
...
+            .toIdentitySet()
+
+        return candidates.filter { candidate ->
+            candidate.label != TEXT_ENTRY_LABEL || candidate in collapsedInputs
+        }
+    }
+
+    private fun List<DetectionResult>.connectedComponents(): List<List<DetectionResult>> {
+        val visited = toIdentitySet()
+        visited.clear()
```
Daniel-ADFA: @jatezzz please check this, this will clear visited before your operations below

sketch-to-ui-plugin/src/main/java/org/appdevforall/codeonthego/computervision/domain/TextInputDetectionCollapser.kt
jatezzz: Thanks for the review! Actually, I think we still need `<= 1` here. Since `textInputs` is a filtered list, `candidates` might not be empty even if `textInputs.size == 0` (e.g., if `candidates` contains 5 items but none of them are `TEXT_ENTRY_LABEL`). If `textInputs` is empty, we still want to early-return the original `candidates` list to avoid running the empty list through the connected components logic.

Daniel-ADFA (approved):

PR #2, opened 2026-04-24
feat/ADFA-3603 ndk plugin

jomen-adfa:
initial commit for ndk plugin

PR #41, opened 2026-07-07
ADFA-4589 | Add suite of AI-powered developer plugins for Code on the Go

jatezzz:
## Description

Adds three new AI plugins and hardens the existing `ai-core` / `ai-assistant` (both already on `main`):

- **Code Suggestions** - inline ghost-text completions as you type.
- **Speech-to-Text** - voice input transcribed into the editor.
- **Vector Search** - embeddings-based semantic search over the project (uses `ai-core`'s `getEmbeddings()`, with a lexical fallback).

All three share `ai-core`'s `LlmInferenceService` (on-device llama.cpp and/or Gemini) instead of bundling their own models.

## Details

Stability & fixes to the existing plugins:
- **ai-core**: reject embedding models on the chat path before they `SIGABRT` the IDE (ADFA-4388); load the user-selected model instead of the newest `.gguf` in Downloads; shut down the `Llm-RunLoop` executor on dispose to stop the classloader leak; move Gemini API key to the `x-goog-api-key` header.
- **ai-assistant**: fetch the live Gemini model catalog (default `gemini-2.5-flash`, retire `gemini-1.5-flash`); fix settings-dialog UX and backend selection/routing.
- Cross-cutting: cancel coroutine scopes on disposal; version ranges, READMEs, in-IDE help HTML, and day/night PNG icons across the AI plugins.

https://github.com/user-attachments/assets/2c5a7dae-a0e5-44b4-b5a9-2bce229d1c39

## Ticket

[ADFA-4589](https://appdevforall.atlassian.net/browse/ADFA-4589)

## Observation

- `llama.cpp` is a submodule at **`ai-core/subprojects/llama.cpp`** (private fork, branch `androidide-custom`, needs SSH): `git submodule update --init ai-core/subprojects/llama.cpp`. A prebuilt `ai-core/libs/v8/llama-v8-release.aar` is committed, so building llama.cpp from source isn't required just to compile.
- `libs/*.jar` are CI-managed and intentionally **not** committed here, but vector-search needs `LlmInferenceService.getEmbeddings()`, so the refreshed `plugin-api.jar` must be in place (CI / `scripts/update-libs.sh`) before building.

Notes:
- Dropped the original .cgp / internal-docs observation: I verified none are tracked in this diff.
- Corrected the submodule path (ai-core/…, not ai-assistant/…) and the "adds four plugins" claim (ai-assistant/ai-core already exist on main; this adds three).

[ADFA-4589]: https://appdevforall.atlassian.net/browse/ADFA-4589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hal-eisen-adfa:
Might be too early to properly review this. They all need version ranges, html documentation, tier 1/2/3 help, and README.md.

CoroutineScope(Dispatchers.IO) and never gets cancel() in dispose().

@Volatile static pluginContext never cleared, native model never unloaded, llama executor never shut down.

ai-assistant has security problems
1. Gemini API key in plaintext SharedPreferences with a misleading "encrypted" comment (AgentSettingsStore.kt:39-40, AiAssistantPlugin.kt:176)
2. API key in URL query string (GeminiBackend.kt:336-341) — should be x-goog-api-key header
3. ai-core makes HTTP calls but doesn't declare network.access/INTERNET.

We also need the new plugin api jar otherwise these won't build.

jatezzz:
@hal-eisen-adfa Thanks for the feedback, I've addressed all the concers, attached the proof that the new plugins build with the new Code On the Go plugin api.

Screenshot 2026-07-16 at 1 17 22 PM

Screenshot 2026-07-16 at 1 47 27 PM

https://github.com/appdevforall/plugin-examples/actions/runs/29523333464

hal-eisen-adfa:
Almost everything is now fixed.

One big gap that wasn't addressed yet: LLamaAndroid needs to expose a shutdown that calls (runLoop as ExecutorCoroutineDispatcher).close() / executor.shutdown(), invoked from AiCorePlugin.dispose().

hal-eisen-adfa (approved):

jatezzz:
Added day and night icons for AI plugins

photo_5102620275676744858_y

photo_5102620275676744859_y

jatezzz:
Update theme and model validation on fresh installs

https://github.com/user-attachments/assets/44726de6-3ac9-4ab1-b255-9139b565bdab

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/LocalLlmBackend.kt
```diff
@@ -333,17 +365,21 @@ class LocalLlmBackend(private val context: PluginContext) : LlmBackend {
...
+        CoroutineScope(Dispatchers.IO).launch {
+            try {
+                unloadModelInternal()
+            } catch (e: Exception) {
+                context.logger.error("Error unloading model during close()", e)
+            } finally {
+                // The Llm-RunLoop executor thread exists even if no model was
+                // ever loaded; shut it down unconditionally so the plugin's
+                // classloader can be collected after unload.
+                llama.shutdown()
```
hal-eisen-adfa, line 407: 🚫 **Blocking (high) — disposing AI Core permanently bricks local inference until an app restart.** `close()` calls `llama.shutdown()` → `runLoop.close()` unconditionally, but `llama` is the process-wide singleton (`LLamaAndroid.instance()` returns `_instance` at `LLamaAndroid.kt:364`, and `runLoop` is a `val` created once at `LLamaAndroid.kt:97`). `close()` runs from `AiCorePlugin.dispose()`, which the CoGo Plugin Manager invokes on disable/uninstall.

After a disable → re-enable within the same process, a fresh `LocalLlmBackend` obtains the same singleton whose executor is now closed → the first `withContext(runLoop)` throws `RejectedExecutionException`, and local inference is dead for the rest of the process lifetime. Either have the singleton lazily recreate its run-loop executor on next use, or don't shut down a process-global run loop from per-plugin `dispose()`.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/AiSettingsFragment.kt
```diff
@@ -387,11 +406,18 @@ class AiSettingsFragment : DialogFragment() {
...
+            // migrate to the first available one so the next chat request
+            // doesn't 404 on a dead model name.
             val currentModel = viewModel.getGeminiModel()
             val currentIndex = models.indexOf(currentModel)
             if (currentIndex >= 0) {
                 modelSpinner.setSelection(currentIndex)
+            } else if (models.isNotEmpty()) {
+                modelSpinner.setSelection(0)
+                val migrated = models[0]
+                viewModel.saveGeminiModel(migrated)
```
hal-eisen-adfa: 🚫 **Blocking — a transient ListModels failure silently overwrites the user's saved model.** When `getGeminiModel()` isn't in `models`, this migrates to `models[0]` and persists it via `saveGeminiModel(...)`. But `models` is frequently `FALLBACK_MODELS` — `fetchGeminiModels()` posts that list on no-key / no-backend / network error — and this observer can't tell a live catalog from the fallback.

So a user who selected e.g. `gemini-2.5-flash-lite` and opens Settings while offline (or when the API errors) is silently switched to `gemini-2.5-flash`, and it's persisted — their choice is lost even after connectivity returns. Only run the migration when the list came from a confirmed live fetch (e.g. pass a flag through, or skip migration when the fallback was used).

vector-search-plugin/src/main/kotlin/com/itsaky/androidide/plugins/vectorsearch/VectorMath.kt
```diff
@@ -0,0 +1,57 @@
...
+     */
+    fun cosineSimilarity(a: FloatArray, b: FloatArray): Float {
+        var dotProduct = 0.0f
+        var normASquared = 0.0f
+        var normBSquared = 0.0f
+
+        // Single-pass calculation for efficiency
+        for (i in a.indices) {
+            val aVal = a[i]
+            val bVal = b[i]
```
hal-eisen-adfa, line 44: 🚫 **Blocking — `ArrayIndexOutOfBoundsException` on mismatched embedding dimensions.** The loop iterates `a.indices` but indexes `b[i]`; if a stored vector is shorter than the query vector, this throws. The index can legitimately hold mixed dimensions — `VectorSearchPlugin.generateEmbedding()` returns the model vector when available else a 384-d lexical fallback — so a model-dim query compared against a 384-d stored vector (or vice-versa) goes out of bounds. It's swallowed by `search()`'s `catch`, so the **entire** query silently returns zero results. Guard with `if (a.size != b.size) return 0f` (or iterate `minOf(a.size, b.size)`).

code-suggestions-plugin/src/main/kotlin/com/itsaky/androidide/plugins/codesuggestions/SuggestionProvider.kt
```diff
@@ -0,0 +1,102 @@
...
+
+private const val TAG = "SuggestionProvider"
+
+/**
+ * Generates code suggestions using an LLM with LRU caching.
+ * Avoids redundant LLM calls for identical context.
+ */
+class SuggestionProvider(private val llmService: LlmInferenceService) {
+
+    private val cache = object : LinkedHashMap<String, String>(16, 0.75f, true) {
```
hal-eisen-adfa: 🚫 **Blocking — shared access-ordered LRU mutated from overlapping coroutines.** `cache` is a `LinkedHashMap(accessOrder = true)` with no synchronization, and `getSuggestion` runs on `Dispatchers.IO`. The debounce `cancel()` can't interrupt the blocking `generateCompletion(...).get()` already in flight, so an old job and a newer one can touch `cache` concurrently — and with `accessOrder = true` even a `get` structurally modifies the map. Fast typing can then trigger `ConcurrentModificationException` / map corruption. Wrap it in `Collections.synchronizedMap(...)` (synchronizing get+put together) or confine cache access to a single thread.

vector-search-plugin/src/main/kotlin/com/itsaky/androidide/plugins/vectorsearch/EmbeddingIndexingService.kt
```diff
@@ -0,0 +1,187 @@
...
+        fun walk(dir: File) {
+            if (files.size >= maxCount) return
+            if (dir.name.startsWith(".") && dir.name != ".") return
+            if (dir.name in skipDirs) return
+
+            try {
+                dir.listFiles()?.forEach { file ->
+                    when {
+                        file.isDirectory -> walk(file)
+                        file.isFile && isCodeFile(file) -> files.add(file)
```
hal-eisen-adfa, line 168: ⚠️ **Blocking (low) — `collectCodeFiles` overshoots `maxCount`.** `return@forEach` is a `continue`, not a break, and this `files.add(file)` branch has no size guard, so a single directory containing more than `maxCount` code files adds all of them — the cap only limits directory *descent* (the `walk` entry guard), not additions within one directory. Add the `files.size >= maxCount` check before `files.add`, or stop the walk properly once the cap is reached.

vector-search-plugin/src/main/kotlin/com/itsaky/androidide/plugins/vectorsearch/VectorSearchPlugin.kt
```diff
@@ -0,0 +1,302 @@
...
+     * Gets the list of code files to index in the current project.
+     */
+    fun getFiles(): List<java.io.File> {
+        val projectDir = java.io.File(
+            System.getProperty("project.dir") ?: System.getProperty("user.dir") ?: return emptyList()
+        )
+        return indexingService.collectFiles(projectDir)
+    }
+
+    override fun searchProject(request: ProjectSearchRequest): CompletableFuture<List<ProjectSearchSection>> {
```
hal-eisen-adfa, line 104: ⚠️ **Blocking (efficiency) — the first search indexes the whole project synchronously.** `searchProject` does `runBlocking { search(...) }`, and `search` → `ensureIndexed` walks up to 500 files, chunks them, and calls a blocking `getEmbeddings(...).get()` once per chunk, serially — re-running whenever `roots` changes. The first query on a fresh project stalls the search worker for a long time (potentially minutes) with no progress or partial results. Prefer indexing in the background on `activate()` and returning partial/empty until the index is ready.

hal-eisen-adfa (changes requested):
**Requesting changes.** Findings from an xhigh code review of the AI-plugin suite that live in this PR's diff, ranked most-severe first. (Two additional items scoped to the stacked API-key PR #49 are posted there.)

Also worth a cleanup pass (non-blocking): `CodeChunker.simpleChunk` has an unreachable "single line too large" `else` branch (the first line is always added, so `currentChunk` is never empty), and `reconcileOverlaps` has an empty `if` body / unused `nextChunk` and just returns its input unchanged.

code-suggestions-plugin/src/main/kotlin/com/itsaky/androidide/plugins/codesuggestions/SuggestionProvider.kt
```diff
@@ -0,0 +1,106 @@
...
+     */
+    suspend fun getSuggestion(
+        fileContent: String,
+        cursorLine: Int,
+        cursorColumn: Int,
+        language: String,
+        prefix: String,
+    ): String = withContext(Dispatchers.IO) {
+        return@withContext try {
+            val cacheKey = "$prefix|$language"
```
hal-eisen-adfa: **Blocking — cache serves wrong completions.** The cache key is `"$prefix|$language"`, but the suggestion is actually generated from `contextBefore` (up to 500 chars before the cursor, built on line 54 and injected into the prompt on line 64). The last word + language is a far coarser key than what the completion depends on.

Failure: type `val` in file A → suggestion S is cached under `val|kotlin`. Later type `val` anywhere else in any Kotlin file → cache hit returns S, even though the surrounding code is completely different.

Fix: include the context in the key, e.g. hash `contextBefore` (or `contextBefore` + `prefix` + `language`) rather than just `prefix|language`.

code-suggestions-plugin/src/main/kotlin/com/itsaky/androidide/plugins/codesuggestions/SuggestionProvider.kt
```diff
@@ -0,0 +1,106 @@
...
+                You are a $language code completion assistant.
+                Complete the following code. Return ONLY the completion text, no explanation.
+
+                Context:
+                $contextBefore
+
+                Complete this:
+            """.trimIndent()
+
+            val config = LlmInferenceService.LlmConfig("local")
```
hal-eisen-adfa: **Blocking — ignores the selected backend.** `LlmConfig("local")` is hardcoded, but `LlmInferenceServiceImpl.generateCompletion` routes on `config.backendId`, and the chat path (`ChatViewModel`) correctly resolves the backend from the `ai_backend_preference`.

Failure: a user who configured only Gemini (no local `.gguf`) gets every request routed to the `local` backend, which has no model → `LlmResponse.failure` → caught → empty string. Suggestions silently never appear. This also contradicts the plugin's own docs ("on-device for Local, or to Google over HTTPS for Gemini").

Fix: read the selected backend from `AgentSettings` (`ai_backend_preference`) the way `ChatViewModel.checkBackendAvailability` does, instead of hardcoding `"local"`.

speech-to-text-plugin/src/main/kotlin/com/itsaky/androidide/plugins/stt/SpeechToTextPlugin.kt
```diff
@@ -0,0 +1,473 @@
...
+
+            // Build a completion prompt for code generation
+            val prompt = """
+                User request: $voiceText
+
+                Generate $language code to fulfill this request. Return only the code, no explanation.
+                Code:
+            """.trimIndent()
+
+            val config = LlmInferenceService.LlmConfig("local")
```
hal-eisen-adfa: **Blocking — ignores the selected backend, and drops the transcript on failure.** Same hardcoded `LlmConfig("local")` as Code Suggestions. Worse: in `handleTranscript` the raw-transcript fallback only runs when `llmService == null`. When AI Core *is* present but local generation fails (e.g. Gemini-only user, no local model), `generateCodeFromVoice` returns `null` → line 278 shows "generation failed" and inserts **nothing** — the recognized speech is lost.

Fix: (a) route to the user-selected backend rather than `"local"`, and (b) when generation returns `null`, fall back to inserting the raw transcript instead of discarding it.

vector-search-plugin/src/main/kotlin/com/itsaky/androidide/plugins/vectorsearch/VectorSearchPlugin.kt
```diff
@@ -0,0 +1,353 @@
...
+        roots: List<File>,
+        llmService: LlmInferenceService?,
+        backendId: String,
+    ): Job? {
+        val rootsKey = roots
+            .map { it.absolutePath }
+            .sorted()
+            .joinToString("|")
+        val running = indexingJob
+        if (running != null && running.isActive) {
```
hal-eisen-adfa, line 203: **Blocking — concurrent search can return another project's results.** `startIndexingIfNeeded` returns the in-flight `indexingJob` whenever it's active, without checking it was launched for the same `rootsKey`.

Failure: a search for `roots=[projectA]` starts indexing; a concurrent search for `roots=[projectB]` sees `running.isActive`, awaits A's job, then reads `getAllEmbeddings()` (project A's data) → returns project A results for a project B query.

Fix: track the `rootsKey` the running job is building and only reuse it when it matches the requested roots; otherwise start (or queue) a build for the new roots.

hal-eisen-adfa (changes requested):
Blocking: 4 correctness issues in the new AI plugin suite. Each is inlined below. Happy to pair on fixes — the two hardcoded-`"local"` ones (#2, #3) and the cache-key one (#1) are the highest impact.

hal-eisen-adfa (approved):

PR #49, opened 2026-07-20
ADFA-4684 | Fix Gemini API key entry: show/hide, trim, and at-rest encryption

jatezzz:
## Description

Fixes the Gemini API key input in the AI Assistant plugin, reported in https://github.com/appdevforall/CodeOnTheGo/issues/1514 where keys wouldn't save correctly and produced 400 "invalid API key" errors. The field now has a working show/hide (eye) toggle that reveals the actual saved key, trims the value on save (dropping the trailing spaces/newlines Gemini rejects), and stops corrupting the key with a masked placeholder on edit. The key is also encrypted at rest with a hardware-backed Android Keystore secret instead of being stored as plaintext.

## Details

Changes:
- **Show/hide toggle**: new `ic_visibility` / `ic_visibility_off` drawables and an `ImageButton` beside the input; toggle swaps `inputType` via an explicit visibility flag (the previous bit-check never flipped back, so it always read "hidden").
- **Edit loads the real key**: Edit now populates the field with the actual saved key (masked) rather than 16 placeholder bullets, so the eye can reveal it and Save can't re-store the placeholder as the key.
- **Trim on save**: `saveButton` trims before persisting (read side already trimmed).
- **Encryption at rest**: `SecureApiKeyStore` (AES/GCM, Android Keystore) in both ai-assistant (write/read) and ai-core (read for API calls); only ciphertext hits SharedPreferences, and legacy plaintext keys migrate transparently on next save.
- **Strings**: hint, content descriptions, and trings.xml`.

https://github.com/user-attachments/assets/db2642c0-6eff-4984-9a36-1b1b4c0a52be

## Ticket

[ADFA-4684](https://appdevforall.atlassian.net/browse/ADFA-4684)

Fixes: 
- https://github.com/appdevforall/CodeOnTheGo/issues/1514
- https://github.com/appdevforall/CodeOnTheGo/issues/1095

## Observation

- The Gemini model list is intentionally left as t (kept for compatibility); retired models can still be selected but now surface a clearer error.
- `SecureApiKeyStore` is duplicated in ai-assistay're separate Gradle projects with no shared module; the two copies **must stay in sync** (same alias `cotg_ai_gemini_key_v1` and transform) or ai-core can't decrypt what ai-assistant wrote. They interoperate because both plugins run in the host app's process (shared Keystore).

[ADFA-4684]: https://appdevforall.atlassian.net/browse/ADFA-4684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jatezzz:
@claude review

jatezzz:
Build Pass

image

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/AiSettingsFragment.kt
```diff
@@ -300,37 +303,69 @@ class AiSettingsFragment : DialogFragment() {
...
+            val timestamp = viewModel.getGeminiApiKeySaveTimestamp()
+            val sdf = SimpleDateFormat("MMMM d, yyyy", Locale.getDefault())
+            val savedDate = sdf.format(Date(timestamp))
+            statusTextView.text = getString(R.string.msg_api_key_saved_on, savedDate)
         }
 
         editButton.setOnClickListener {
             updateUiState(isEditing = true)
-            apiKeyInput.setText("••••••••••••••••")
+            apiKeyInput.setText(viewModel.getGeminiApiKey().orEmpty())
```
hal-eisen-adfa: 🚫 **Blocking — Keystore decryption on the main thread.** This click handler calls `getGeminiApiKey()`, which runs `SecureApiKeyStore.decrypt()` (AndroidKeyStore `getEntry` + AES/GCM `Cipher.doFinal`) synchronously on the UI thread. For a previously-encrypted key that's real keystore IPC plus crypto on the main thread — a jank / `StrictMode` risk on the Edit tap. Load the key into the field from a coroutine on `Dispatchers.IO` (as the save path already does) and set the text on the main thread once it resolves.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/FilePickerDialogFragment.kt
```diff
@@ -1,143 +1,232 @@
...
         }
 
-        // Add directories and files
-        files.forEach { file ->
-            val prefix = if (file.isDirectory) "📁 " else "📄 "
-            val suffix = if (selectedFiles.contains(file)) " ✓" else ""
-            items.add("$prefix${file.name}$suffix")
-            fileList.add(file)
-        }
+        directory.listFiles()
```
hal-eisen-adfa, line 194: 🚫 **Blocking (defense-in-depth) — root confinement can be escaped through a symlink.** `computeListing` enumerates `directory.listFiles()` and adds every child as a navigable row, but `isWithinRoot` is only enforced on the `..` (upward) entry above. If the project root contains a directory symlink pointing outside it (e.g. to `/storage/emulated/0`), descending into that entry lists files outside the confined root — defeating the "can't be used to reach arbitrary files on the device" guarantee this class documents. Resolve each listed directory's canonical path and skip (or refuse to `populate`) any child that falls outside `rootDirectory`.

hal-eisen-adfa (changes requested):
**Requesting changes.** Two blocking items scoped to this PR's diff (the API-key security + FilePicker rewrite). The larger AI-plugin findings from the review belong to the base PR #41 and are posted there, not here.

The rest of the PR is solid: AES/GCM at-rest encryption with legacy-plaintext passthrough, `inputType="textPassword"` on the key field, save-failure surfaced instead of crashing, and the in-place FilePicker refresh that fixes the second-"Toggle All" crash.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/GeminiBackend.kt
```diff
@@ -50,38 +64,69 @@ class GeminiBackend(private val context: PluginContext) : LlmBackend {
...
-            context.logger.error("GeminiBackend: Error getting preferences", e)
-            null
+        val stored = agentPrefs()?.getString(KEY_API_KEY, null)
+        if (stored.isNullOrBlank()) {
+            keyCache = null
+            return null
         }
-        return prefs?.getString("gemini_api_key", null)?.trim()?.takeIf { it.isNotBlank() }
+        keyCache?.let { (raw, plain) -> if (raw == stored) return plain }
+        if (Looper.myLooper() == Looper.getMainLooper()) {
```
hal-eisen-adfa, line 98: **[Medium] `isAvailable()` can spuriously report "no key" on a cold main-thread call.**

On a main-thread cache *miss* this returns `null` and only schedules a background refresh, so the synchronous `LlmBackend.isAvailable()` returns `false` even when a decryptable key is stored. Pre-PR, `readGeminiApiKey()` read the plaintext synchronously and worked on any thread — so this is a behavior regression that can surface as "Gemini backend not available" the first time availability is checked from the UI thread before any off-thread read warms the cache.

The doc comment calls this "worst case reports no key once," which is fair, but a caller gating UI on `isAvailable()` would see the wrong answer at exactly the wrong moment. Consider warming the cache during `initialize()/activate()`, or documenting the ordering requirement on the interface.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/security/SecureApiKeyStore.kt
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * The alias and transform below are mirrored verbatim in ai-core's
+ * `SecureApiKeyStore` so a key written here can be decrypted there — both
+ * plugins run in the host app's process (same UID) and therefore share one
+ * Android Keystore. Keep the two copies in sync.
+ */
+object SecureApiKeyStore {
+    private const val TAG = "SecureApiKeyStore"
+    private const val KEYSTORE = "AndroidKeyStore"
+    private const val ALIAS = "cotg_ai_gemini_key_v1"
```
hal-eisen-adfa, line 31: **[Medium] `SecureApiKeyStore` is duplicated byte-for-byte in `ai-assistant` and `ai-core`, kept in sync only by a comment.**

If `ALIAS`, `TRANSFORM`, `IV_LEN`, or `ENC_PREFIX` ever drift between the two copies, the build stays green but `ai-core` silently fails to decrypt the key `ai-assistant` wrote → "backend not available" with nothing pointing at the cause. The class doc acknowledges this, but the constants remain unenforced across the file boundary.

If a shared module isn't feasible given the per-plugin `.cgp` packaging, a cheap mitigation is a unit test asserting the constant values (or a checked-in hash of them) so drift fails CI rather than a device.

hal-eisen-adfa (commented):
Code review of the Gemini API-key security work. The crypto core is sound — AES/GCM with a fresh random 12-byte IV per encrypt, GCM tag verification, fail-closed decrypt, and a sensible legacy-plaintext migration. No blockers; 6 non-critical findings inline below, ranked by severity. Note this is a static review — the encryption, on-device migration, and FLAG_SECURE screenshot-blocking still need device verification (install both .cgp, save a key, confirm ciphertext in prefs, revoke the Keystore alias to exercise the 'unreadable' path).

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/AiSettingsFragment.kt
```diff
@@ -318,47 +333,102 @@ class AiSettingsFragment : DialogFragment() {
...
-            apiKeyInput.setText("")
-        } else {
-            updateUiState(isEditing = false)
-            val timestamp = viewModel.getGeminiApiKeySaveTimestamp()
-            if (timestamp > 0) {
-                val sdf = SimpleDateFormat("MMMM d, yyyy", Locale.getDefault())
-                val savedDate = sdf.format(Date(timestamp))
-                statusTextView.text = "API Key saved on: $savedDate"
+        viewLifecycleOwner.lifecycleScope.launch {
+            val savedApiKey = viewModel.getGeminiApiKey()
```
hal-eisen-adfa, line 337: **[Low-Med] Initial load treats a present-but-undecryptable key as "no key", silently.**

When the Keystore entry is lost (app data cleared, OEM drops the alias), `getGeminiApiKey()` returns `null` here → `hasKey = false` → an empty editable field, implying nothing was ever saved. The **Edit** button path shows `msg_api_key_unreadable` for the identical state; this initial load is silent. Consider distinguishing "nothing stored" from "stored but undecryptable" here too (e.g. peek at the raw pref) and showing the same warning, so the user isn't misled about whether a key exists.

ai-assistant/src/main/AndroidManifest.xml
```diff
@@ -40,9 +40,10 @@
             android:name="plugin.editor_tabs"
             android:value="1" />
 
+        <!-- ide.settings: AiAssistantPlugin's migration reads the host store. No network.access: ai-core makes the Gemini calls. -->
         <meta-data
             android:name="plugin.permissions"
-            android:value="filesystem.read,filesystem.write,system.commands,project.structure" />
+            android:value="filesystem.read,filesystem.write,system.commands,project.structure,ide.settings" />
```
hal-eisen-adfa: **[Low-Med] New `ide.settings` permission — justification comment doesn't match the code.**

The comment above says the migration "reads the host store," but the migration reads this plugin's *private* `AgentSettings` via `getPluginSharedPreferences("AgentSettings")`, and `ai-core` reads the same prefs through `SharedServices` without declaring `ide.settings`. So either the permission is unnecessary (drop it, for least privilege) or the rationale is wrong (fix the comment to state what actually needs it). Worth confirming against the host's permission gating before merge.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/viewmodel/AiSettingsViewModel.kt
```diff
@@ -188,21 +193,39 @@ class AiSettingsViewModel(
...
     }
 
-    fun getGeminiApiKey(): String? {
-        return getPluginPrefs()?.getString("gemini_api_key", null)
+    /**
+     * Decrypt the stored key off the main thread (Keystore IPC + AES/GCM), upgrading a
+     * pre-encryption plaintext key to ciphertext in passing so existing installs actually
+     * end up encrypted rather than waiting for the user to re-enter the key.
+     */
+    suspend fun getGeminiApiKey(): String? = withContext(ioDispatcher) {
```
hal-eisen-adfa, line 237: **[Low] Trim handling is inconsistent between save and read/migrate.**

`saveGeminiApiKey()` trims before encrypting, but `SecureApiKeyStore.readAndMigrate()` re-encrypts a *legacy plaintext* value verbatim (untrimmed) and this `getGeminiApiKey()` returns it untrimmed — so the Edit field can show a whitespace-padded key. Only the two backends `.trim()` on read, so the persisted/displayed value disagrees with what's actually sent. Trimming in the migration path (and/or here) would make the three paths agree.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/LocalLlmBackend.kt
```diff
@@ -32,6 +35,15 @@ class LocalLlmBackend(private val context: PluginContext) : LlmBackend {
     private val llama by lazy { LLamaAndroid.instance() }
     private val scope = CoroutineScope(Dispatchers.IO)
 
+    /**
+     * Separate from [scope] because [close] cancels [scope] and then has to run the unload —
+     * work submitted to a cancelled scope never starts.
+     */
+    private val teardownScope = CoroutineScope(Dispatchers.IO)
```
hal-eisen-adfa: **[Low] `teardownScope` is created but never cancelled — it leaks for the plugin's lifetime.**

`close()` cancels `scope` and launches teardown on this separate `teardownScope`, which is never cancelled anywhere. On repeated load/unload cycles each backend instance leaves a live scope behind. Cancel `teardownScope` after the teardown job joins (e.g. at the end of `awaitClose`, or from the job's completion handler).

hal-eisen-adfa (approved):

PR #50, opened 2026-07-22
ADFA-2707 | Model-driven open-file tool + reliable local-LLM agent loop

jatezzz:
## Description

Adds an `open_file` tool to the AI Assistant plugin that exercises the real agent path (model requests a tool → agent loop executes it → file opens in the IDE editor), and hardens the text-based agent loop so it behaves reliably on
weak local models.

The core problem this fixes: path resolution previously relied on the `project.dir` system property (never set) and fell back to `user.dir`, which on Android is `"/"`. Every relative path (e.g. `.gitignore`) resolved against the filesystem root and was rejected as "outside the project directory," so the tool never worked. Resolution is now anchored to the actual open project, with basename/slash-prefixed fallbacks so a hallucinated bare filename still lands on the right file, or reports an honest, disambiguable failure.

## Details

**`open_file` tool + path resolution (`PathGuard`, `OpenFileHandler`)**
- Project root now sourced (in priority order) from a host-backed provider wired at plugin activation to `IdeProjectService.currentProject.rootDir`, queried lazily so a mid-session project switch is picked up automatically; legacy system-property fallbacks retained.
- New `PathGuard.resolve()` returns a sealed `Resolution` (`Resolved` / `Ambiguous` / `NotFound` / `Escaped`) so handlers can act or explain the miss.
- `findByName()` resolves a bare filename to its real path by walking the project root (case-insensitive), skipping generated/noise dirs (`build`, `.git`, `.gradle`, `.idea`, `node_modules`, `.cxx`) and hidden dirs, with a result cap.
- `OpenFileHandler` opens the editor tab on the main dispatcher (`execute()` runs on IO) and surfaces ambiguous matches back to the model instead of a stack trace.

**Reliable local-LLM agent loop (`AgentLoop`, `ChatViewModel`)**
- Anti-fabrication prompting, honest failure reporting, stop-after-success, and repeated-call / max-iteration backstops to keep weak models from looping.
- Replaces the old `ToolCallExtractor` with a dedicated `AgentLoop`.

**Streaming cancellation + crash fixes (`ai-core`)**
- Cancellable streaming for both Local and Gemini backends (`CancellableBackend`, `LLamaAndroid`, `llama-android.cpp`), plus IDE-crash fixes.

**Tests**
- New JVM unit tests: `AgentLoopTest`, `ExecutorTest`, `ToolCallExtractorTest`, `ListFilesHandlerTest`, `OpenFileHandlerTest`, `PathGuardTest` (~1,000 lines of test coverage).

https://github.com/user-attachments/assets/2664c68a-d9e5-4714-85c1-e5d7f5458daf

## Ticket

[ADFA-2707](https://appdevforall.atlassian.net/browse/ADFA-2707)

## Observation

- `ai-core/libs/v8/llama-v8-release.aar` and `llama-api.jar` are rebuilt binaries (native cancellation support), regenerated via `ai-core/scripts/rebuild-llama-aar.sh`.

[ADFA-2707]: https://appdevforall.atlassian.net/browse/ADFA-2707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hal-eisen-adfa (commented):
## Code review — 16 findings (correctness-focused, xhigh effort)

Reviewed `68a6f98` against the PR base `refactor/ADFA-4589-ai-hackathon-plugin` (not `main`) across 10 finder angles plus direct source verification. Inline comments below, ranked roughly by severity.

**Gate-worthy:** (1) PathGuard sandbox regresses to accept-all when the project root falls back to `/`; (2) the local-model GBNF grammar makes 7 registered tools structurally uncallable; (3) Clear-Chat history race; (4) session-wide `lastToolResult` mislabels replies as "action failed"; (5) `open_file` arg-remap gap; (6) grammar permits raw newlines → org.json drops the call.

Then a concurrency/cancellation tier (StringBuilder streaming race, no single-flight on the `LLamaAndroid` singleton, Gemini/local cancellation not actually aborting native/HTTP work) and agent-loop edge cases (co-emitted `respond`, `REPEATED` false-abort), plus JNI null-checks.

**Verified NOT problems:** the JNI/API boundary is consistent (all `new_grammar_sampler`/`setGrammar` signatures, `State.Loaded.model`, `LlmConfig.extraParams`, `ToolResult` fields, `R.string.*` all line up); the adapter/fragment UX issues some reviewers might flag are on unchanged lines outside this diff. No CLAUDE.md convention violation.

Two comments (`Executor.kt`, `ToolCallExtractor.kt`) anchor to the nearest in-diff line in the same function; the true line is noted in the body.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/tool/handlers/PathGuard.kt
```diff
@@ -30,13 +73,87 @@ object PathGuard {
...
-        val root = File(projectRoot()).canonicalPath
-        val file = if (path.startsWith("/")) File(path) else File(projectRoot(), path)
+        val rootFile = File(projectRoot())
+        val root = rootFile.canonicalPath
+        val file = if (path.startsWith("/")) File(path) else File(rootFile, path)
         val canonical = file.canonicalPath
-        val contained = canonical == root || canonical.startsWith(root + File.separator)
+
+        val rootWithSep = if (root.endsWith(File.separator)) root else root + File.separator
+        val contained = canonical == root || canonical.startsWith(rootWithSep)
```
hal-eisen-adfa, line 79: **🔴 Security — filesystem sandbox regresses to accept-all when the root is `/`.**

This PR added the `rootWithSep` trailing-separator normalization (good — it fixes the sibling-dir bug where root `/a/proj` would match `/a/project2`), but it regressed the `root == "/"` case:

- **Base:** `canonical.startsWith(root + File.separator)` → `startsWith("//")` → `"/etc/passwd"` is **rejected**.
- **This PR:** `root` already ends in `/`, so `rootWithSep` stays `"/"` → `canonical.startsWith("/")` is **true for every absolute path**.

`projectRoot()` falls back to `System.getProperty("user.dir")`, which is `"/"` on Android, and the `IdeProjectService`-backed provider returns `null` when no project is open or it throws (the `catch` swallows to null). So `read_file`/`open_file`/`list_files` can escape to arbitrary absolute paths (`/data/data/<other-app>/...`, `/system/...`).

**Fix:** treat a root of `"/"` (or blank/non-existent) as "no valid root" and reject, rather than accepting everything.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/viewmodel/ChatViewModel.kt
```diff
@@ -51,6 +57,31 @@ class ChatViewModel(
...
+        /**
+         * GBNF forcing a weak local model to emit one well-formed `<tool_call>`;
+         * passed to the local backend via [EXTRA_PARAM_GRAMMAR] (cloud ignores it).
+         * `args` may be empty or hold multiple `"key":"value"` pairs; values are
+         * JSON strings (newlines allowed). Keep the `tool` alternatives in sync with
+         * the exposed handlers. `~` stands in for an escaped quote, substituted below.
+         */
+        private val localToolCallGrammar: String = """
+            root ::= "<tool_call>{~tool~:" tool ",~args~:" args "}</tool_call>"
+            tool ::= "~open_file~" | "~read_file~" | "~list_files~" | "~search_project~" | "~$RESPOND_TOOL~"
```
hal-eisen-adfa: **🔴 Capability loss — the local-model grammar makes 7 registered tools uncallable.**

`tool ::=` whitelists only `open_file | read_file | list_files | search_project | respond`, but the router registers `create_file`, `update_file`, `add_dependency`, `run_app`, `gradle_sync`, `read_build_output`, `generate_from_template` (lines 170–183), and `buildSystemPrompt()` advertises all of them. On the local backend the GBNF grammar is a hard token mask, so the model **physically cannot emit** those tokens — "create MainActivity" / "run the app" silently degrade to read/list/search/respond. Cloud (Gemini, ignores the grammar) is unaffected, so this is invisible from a clean build.

The comment on line 71 ("Keep the `tool` alternatives in sync with the exposed handlers") admits the manual coupling. **Fix:** generate the `tool ::=` alternatives from `toolRouter.getAllHandlers()` so the grammar and the prompt can't drift.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/viewmodel/ChatViewModel.kt
```diff
@@ -51,6 +57,31 @@ class ChatViewModel(
...
+            args ::= "{}" | "{" pair ("," pair)* "}"
+            pair ::= "~" key "~:~" val "~"
+            key  ::= [a-z_]+
+            val  ::= char*
+            char ::= [^~\\] | "\\" [~\\/bfnrt]
```
hal-eisen-adfa: **🟠 Correctness — the grammar permits raw newlines inside JSON strings, so org.json rejects the tool call.**

After the `~`→`\"` substitution, `char ::= [^"\] | "\" ["\/bfnrt]`. The `[^"\]` class matches **any byte except quote/backslash, including a literal newline (0x0A)** and other control chars. GBNF allows it, but a raw control char inside a JSON string is invalid JSON — `JSONObject` in `ToolCallExtractor.parseToolJson` throws, the call is dropped, `extractToolCalls` returns empty, and `AgentLoop` ends `COMPLETED` with "(No response)". The doc comment "newlines allowed" (line 71) is exactly the bug: JSON needs `\n`, not a literal break.

**Fix:** exclude control chars from the unescaped class, e.g. `char ::= [^"\\u0000-\u001F] | "\" ["\/bfnrt]` (or `[ -!#-[]-~...]`), forcing newlines through the `\n` escape.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/viewmodel/ChatViewModel.kt
```diff
@@ -509,176 +559,260 @@ class ChatViewModel(
...
+                            override suspend fun onRepeatedToolCalls(turns: Int) {
+                                addSystemMessage(
+                                    str(R.string.agent_repeated_calls),
+                                    MessageStatus.SENT
+                                )
                             }
                         }
+                    )
+                } finally {
+                    _history.value = history.toList()
```
hal-eisen-adfa: **🟠 Race — Clear Chat during an in-flight run leaves the "cleared" history intact.**

`history` is captured at line 601. If the user taps Clear Chat mid-run, `clearMessages()` cancels the job and sets `_history.value = emptyList()` (line 815). But cancellation unwinds the suspended coroutine *after* `clearMessages()` returns, and this `finally` runs `_history.value = history.toList()` — overwriting the empty history with the pre-clear transcript. The next message is then sent with the context the user thought they cleared.

**Fix:** guard the restore with the epoch (only restore if `generationEpoch.get() == epoch`), or don't persist history from a cancelled run.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/viewmodel/ChatViewModel.kt
```diff
@@ -509,176 +559,260 @@ class ChatViewModel(
...
+                        // Already cancelled; the awaiting loop was unblocked by job cancel.
+                        deferred.complete(response.text)
+                        return
+                    }
+                    val durationMs = System.currentTimeMillis() - startTime
+                    val toolCalls = ToolCallExtractor.extractToolCalls(response.text)
+                    val respondCall = toolCalls.firstOrNull { it.name == RESPOND_TOOL }
+                    val respondMessage = respondCall?.args?.get("message")?.toString()
+
+                    val lastToolResult = _messages.value.lastOrNull { it.sender == Sender.TOOL }
```
hal-eisen-adfa: **🟠 Correctness — `lastToolResult` scans the whole session, so a plain reply is mislabeled "action failed".**

`_messages.value.lastOrNull { it.sender == Sender.TOOL }` picks the last tool message across the **entire session**, not the current turn. If any earlier tool failed (an `ERROR` `Sender.TOOL` message persists), a later unrelated `respond` (e.g. the user says "Hi") hits the `respondCall != null && lastToolFailed` branch (line 716) and the greeting is replaced with the generic "action failed" string.

**Fix:** scope `lastToolFailed` to the tool results produced in *this* agent-loop turn (e.g. pass the current turn's results into the render), not a session-wide scan.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/viewmodel/ChatViewModel.kt
```diff
@@ -509,176 +559,260 @@ class ChatViewModel(
...
+            )
+            _messages.value = _messages.value + agentMessage
+            syncMessageToSession(agentMessage)
+        }
+
+        try {
+            llmService.generateStreaming(prompt, config, object : LlmInferenceService.StreamCallback {
+                override fun onToken(token: String) {
+                    if (isStale()) return  // Stop pressed — ignore late tokens.
+                    responseBuilder.append(token)
```
hal-eisen-adfa, line 695: **🟠 Data race — `responseBuilder` (StringBuilder) is written on the backend thread and read on Main concurrently.**

`onToken` runs `responseBuilder.append(token)` on the backend/IO streaming thread, while a previously-launched `viewModelScope.launch(Dispatchers.Main)` reads `responseBuilder.toString()` (line 692). `StringBuilder` is not thread-safe — concurrent `append`+`toString` can throw `ArrayIndexOutOfBounds`/`NegativeArraySize` or produce a garbled bubble mid-stream.

**Fix:** snapshot the string on the producer thread and pass the immutable value into the Main coroutine, or guard with a lock / use a thread-safe buffer.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/viewmodel/ChatViewModel.kt
```diff
@@ -509,176 +559,260 @@ class ChatViewModel(
             return
         }
 
+        // Guard against re-entry: an in-flight run must be stopped (Stop button)
+        // before a new one starts, so the ViewModel is correct regardless of the
+        // UI. Otherwise the old loop keeps executing tools and a second concurrent
+        // generation races the first on the single-threaded backend.
+        if (generationJob?.isActive == true) {
```
hal-eisen-adfa: **🟡 Concurrency — post-Stop re-entry window lets a second generation race the backend.**

`stopProcessing()` sets `generationJob = null` (line 862) and calls `cancelGeneration()`, but the native inference may still be draining. This guard only checks `generationJob?.isActive`, which is now `null`, so an immediate re-send proceeds and launches a fresh `generateStreaming` while the previous llama.cpp run is still finishing on the single-threaded backend — the exact hazard the guard comment warns about. The epoch guard suppresses only stale *UI* writes, not backend contention.

**Fix:** gate re-entry on backend idle (or an explicit "generating" flag cleared when the native run actually completes), not on the nulled job reference.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/tool/AgentLoop.kt
```diff
@@ -0,0 +1,178 @@
...
+            val calls = extractToolCalls(text)
+            if (calls.isEmpty()) {
+                return Result(turn, StopReason.COMPLETED)
+            }
+
+            // Terminal tool: the model signals it's finished by calling it (with a
+            // `message`). Emit that as the final answer and stop — don't dispatch
+            // it as a real tool (there's no handler for it).
+            terminalTool?.let { tt ->
+                calls.firstOrNull { it.name == tt }?.let { done ->
```
hal-eisen-adfa: **🟡 Logic — a turn co-emitting `respond` + a real tool drops the real tool.**

`calls.firstOrNull { it.name == tt }` returns `COMPLETED` as soon as it sees the terminal tool, so if an unconstrained backend (Gemini) emits `<tool_call>open_file…</tool_call><tool_call>respond…</tool_call>` in one turn, `open_file` is never dispatched — the user is told the task finished while the action never ran.

**Fix:** only treat `respond` as terminal when it's the *sole* call in the batch; otherwise execute the real tools first (or ignore `respond` when other calls are present).

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/tool/AgentLoop.kt
```diff
@@ -0,0 +1,178 @@
...
+                    events.onFinalAnswer(turn, done.args["message"]?.toString().orEmpty())
+                    return Result(turn, StopReason.COMPLETED)
+                }
+            }
+
+            // Stuck-detection: a weak model can loop by re-requesting the exact
+            // same tool calls forever. If it does, stop instead of spinning.
+            val signature = signatureOf(calls)
+            if (signature == previousSignature) {
```
hal-eisen-adfa, line 127: **🟡 Logic — stuck-detection aborts legitimate identical retries.**

A `read_file` result capped at `toolOutputCharLimit` (4000) shows `…[truncated]`; the model rationally re-issues the *same* `read_file`. `signatureOf()` matches `previousSignature` and the run aborts with `REPEATED`, even though the retry was valid.

**Fix:** only treat a repeat as stuck when the prior result was successful/unchanged, or allow N identical repeats before bailing, or vary the feedback (e.g. offer a paging offset) so the re-request isn't byte-identical.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/tool/handlers/PathGuard.kt
```diff
@@ -30,13 +73,87 @@ object PathGuard {
...
+     */
+    fun resolve(path: String): Resolution {
+        val primary = resolveWithin(path)
+        val relative = if (path.startsWith("/")) resolveWithin(path.removePrefix("/")) else null
+
+        (primary?.takeIf { it.exists() } ?: relative?.takeIf { it.exists() })?.let {
+            return Resolution.Resolved(it)
+        }
+
+        val matches = findByName(path)
```
hal-eisen-adfa, line 161: **🟡 Security — `resolve()` returns a `findByName` basename match without re-checking containment.**

Two issues:
1. **Silent substitution:** `read_file("/etc/config.xml")` → primary escapes, relative `etc/config.xml` doesn't exist, then `findByName("config.xml")` returns *any* project file named `config.xml` as `Resolved` — the caller is told the absolute path was read but a different file was served.
2. **Symlink escape:** `findByName`'s `onEnter` (line 108) filters directories by name only, with no `Files.isSymbolicLink` check, so `walkTopDown` follows a symlinked project subdir and can match files physically outside the canonical root — and those matches bypass `resolveWithin` entirely.

**Fix:** re-run the containment check on each `findByName` result, and skip symlinked directories during the walk.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/LocalLlmBackend.kt
```diff
@@ -298,28 +316,45 @@ class LocalLlmBackend(private val context: PluginContext) : LlmBackend {
                 var tokenCount = 0
                 val responseBuilder = StringBuilder()
 
+                // Apply the caller-supplied grammar (cleared in finally); relies
+                // on the single-threaded native run loop serializing generations.
+                val grammar = config.extraParams?.get(EXTRA_PARAM_GRAMMAR) as? String
+                llama.setGrammar(grammar?.takeIf { it.isNotBlank() })
```
hal-eisen-adfa: **🟡 Concurrency — grammar + generation state live on the singleton `LLamaAndroid` with no single-flight guard.**

`generateStreaming` sets `llama.setGrammar(grammar)` here then `send()`; the non-streaming `generate()` path sets `setGrammar(null)`. The `@Volatile grammar` is read later on the runLoop thread, so a concurrent `generate()` can null it before the streaming `send()` reads it → the tool-call turn runs unconstrained and emits malformed calls. `flowOn(runLoop)` serializes individual dispatched blocks, **not whole flows** (each `emit` is a suspension/hand-off), so two concurrent generations can also interleave `completion_loop` calls on the shared `State.Loaded` context/batch/kv-cache → garbled output or native SIGABRT. The "single-threaded run loop serializes generations" comment doesn't hold.

**Fix:** add a real single-flight guard (mutex) around a whole generation in the backend, and bind the grammar to the specific `send()` call rather than shared singleton state.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/GeminiBackend.kt
```diff
@@ -184,15 +184,11 @@ class GeminiBackend(private val context: PluginContext) : LlmBackend {
                 var chunkCount = 0
 
                 try {
-                    // Iterate through stream chunks as they arrive
                     for (response in responseStream) {
```
hal-eisen-adfa: **🟡 Cancellation — Gemini Stop never aborts the blocking stream.**

`cancelStreaming()` cancels the coroutine, but this `for (response in responseStream)` loop has no suspension point / `ensureActive()`, so cancellation isn't observed — the iterator keeps pulling chunks and firing `callback.onToken()` for the full server response, and `responseStream.close()` only runs in the `finally` after natural completion (leaking the connection and updating the UI after Stop).

**Fix:** check `coroutineContext.ensureActive()` (or `isActive`) each iteration and close the stream on cancel.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/LlmInferenceServiceImpl.kt
```diff
@@ -130,5 +130,8 @@ class LlmInferenceServiceImpl : LlmInferenceService {
     override fun cancelGeneration() {
         currentGeneration?.cancel(true)
```
hal-eisen-adfa, line 173: **🟡 Cancellation — `CompletableFuture.cancel(true)` can't stop the backend coroutine.**

Backends finish their future via `future.complete(...)`; no thread is bound to it and `CompletableFuture` ignores `mayInterruptIfRunning`, so `currentGeneration?.cancel(true)` only flips the future's state. Only the streaming path (via `CancellableBackend.cancelStreaming()`) actually stops — non-streaming `generate()`/`generateWithHistory()` keep running to completion after Stop, and the local non-streaming path never sets `currentStreamingJob` and has no `ensureActive()` in its collect, so it ties up the single Llm-RunLoop thread up to `maxTokens`.

**Fix:** make the non-streaming paths cancellable too (track their Job and cancel it, add `ensureActive()` in the collect loop), and/or signal the native stop flag.

ai-core/llama-impl/src/main/cpp/llama-android.cpp
```diff
@@ -554,6 +554,43 @@ Java_android_llama_cpp_LLamaAndroid_new_1sampler(JNIEnv *, jobject) {
...
+ */
+extern "C"
+JNIEXPORT jlong JNICALL
+Java_android_llama_cpp_LLamaAndroid_new_1grammar_1sampler(
+        JNIEnv *env, jobject, jlong model_pointer, jstring grammar) {
+    const auto model = reinterpret_cast<llama_model *>(model_pointer);
+    if (model == nullptr) return 0;
+    const llama_vocab *vocab = llama_model_get_vocab(model);
+
+    const char *grammar_cstr = env->GetStringUTFChars(grammar, nullptr);
```
hal-eisen-adfa, line 600: **🟡 JNI robustness — `GetStringUTFChars` (and `llama_model_get_vocab`) results aren't null-checked, turning OOM/invalid input into a native crash instead of the documented fallback.**

`env->GetStringUTFChars(grammar, nullptr)` returns `NULL` (raising a pending `OutOfMemoryError`) under memory pressure; it's passed straight into `llama_sampler_init_grammar`, which `strlen`/derefs `NULL` → SIGSEGV, then `ReleaseStringUTFChars(NULL)`. There's also no `env->ExceptionCheck()`, so later JNI calls run with a pending exception. The function's own KDoc promises it returns 0 for graceful fallback.

**Fix:** `if (grammar == nullptr) return 0;`, check `grammar_cstr` for null after `GetStringUTFChars` (return 0), and null-check `vocab` before use.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/tool/Executor.kt
```diff
@@ -107,13 +108,7 @@ class Executor(
             return ToolResult.failure(message)
```
hal-eisen-adfa, line 110: **🟡 Correctness — `open_file` (new in this PR) fails on a `path` arg that `read_file` tolerates.** _(the defect is the remap at line 93, shown just above.)_

The `"path" → "file_path"` normalization at line 93 is hardcoded to `toolName == "read_file"`, but `open_file` also requires `file_path` (`requiredArgsForTool`). A model emitting `{"tool":"open_file","args":{"path":"MainActivity.java"}}` (it uses `path`/`file_path` interchangeably) gets a spurious "Missing required argument(s): file_path" here — the identical `read_file` call would have been remapped and succeeded.

**Fix:** apply the `path`→`file_path` remap for any tool whose required args include `file_path` (or normalize arg aliases centrally), not just `read_file`.

ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/tool/ToolCallExtractor.kt
```diff
@@ -123,56 +117,6 @@ class ToolCallExtractor {
             return toolCalls
```
hal-eisen-adfa, line 131: **🟢 Correctness (pre-existing, but this PR leans on it harder) — the bare-JSON brace counter doesn't skip string literals.** _(the counter is at line 85, in this `extractFromJsonObjects`.)_

`extractFromJsonObjects` counts `{`/`}` structurally without ignoring braces inside string values, so a tool arg whose value contains an unbalanced brace (e.g. `create_file`/`update_file` content like `"if (x) {"`) mis-balances the counter, the extracted substring is malformed JSON, `parseToolJson` returns null, and the call is silently dropped. This is the Strategy-2 fallback (only runs when no `<tool_call>` XML is found), and removing Strategy 3 in this PR makes the JSON path more load-bearing.

**Fix:** track in-string/escape state while scanning braces, or extract candidate objects with a real JSON tokenizer.

hal-eisen-adfa:
### Plugin review finding — reflection into ai-core internals (rubric 6.5)

`AiSettingsViewModel.fetchGeminiModels()` reaches the Gemini backend's model list reflectively:

```kotlin
// AiSettingsViewModel.kt:258-259
val method = geminiBackend.javaClass.getMethod("listModels")
val futureResult = method.invoke(geminiBackend)
```

`geminiBackend` comes from `llmService.getBackend("gemini")`, an ai-core (`com.itsaky.androidide.*`) type, and `listModels()` isn't on the shared `LlmInferenceService`/`LlmBackend` interface — so it's invoked reflectively across the plugin classloader boundary. This trips submission rubric clause **6.5 (no reflection into IDE internals)**.

It's moderate severity — `getMethod` only reaches a public method, with no `setAccessible`/private-field access — but it's still reflection into a sibling-plugin internal and is brittle (a rename in ai-core breaks it silently at runtime rather than at compile time).

**Suggested fix:** add `listModels()` to the shared `LlmInferenceService` (or `LlmBackend`) interface exposed by `plugin-api.jar`, then call it directly and drop the `getMethod`/`invoke`. This removes the reflection and makes the contract compile-checked.

*(Note: this file isn't part of the current PR diff, so posting as a general comment rather than inline.)*

hal-eisen-adfa (approved):

PR #56, opened 2026-07-31
ADFA-4857 | Add project-to-template plugin

hal-eisen-adfa:
## Summary
Incorporates the standalone `templatize-project` plugin into the monorepo as **`project-to-template/`**, conforming to repo conventions and wired into the build + website-deploy pipelines.

The plugin adds a **Project to Template** sidebar item that converts the currently-open Android project into a Code On The Go `.cgt` template bundle (original untouched — it copies then tokenizes with Pebble `${{TOKEN}}`s) and can install it straight into the New Project template picker via `IdeTemplateService`.

## What changed on the way in
- **Full rename** to `org.appdevforall.projecttotemplate` / "Project to Template".
- **Shared infra**: uses repo-root `../libs/*.jar` and `../gradlew` — no bundled jars or wrapper.
- **Tooltip fix**: `getTooltipCategory()` now returns the full `plugin_org.appdevforall.projecttotemplate` id, so the in-IDE tooltip renders real text instead of the literal `n/a`.
- **Security**: sanitizes the user-typed template name before using it as a filesystem path.
- **Docs**: Tier 3 offline help (`assets/docs/index.html`) + top-level website `project-to-template-documentation.html`.
- **New icon**: solid project card → arrow → dashed template card (day/night PNGs + monochrome `ic_plugin` vector), replacing the 4-square grid glyph. Source SVGs + generator under `icon-src/`.
- **Text**: normalized user-facing strings to "Code On The Go".
- **Wiring**: README Examples row + `update-libs.yml` website MAP entry.

## Verification
- `../gradlew assemblePlugin` and `assemblePluginDebug` both build clean.
- Passed the `plugin-review` rubric (green-light; 6.1 min-only IDE version is the repo norm).
- **Device-verified on emulator** end-to-end: installs, activates (survives restart), sidebar + 3-tier tooltip render (no `n/a`), and a real conversion of an open project (15 modified) produces a `.cgt` that installs via `IdeTemplateService` (`[OK] Installed template`).


project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/fragments/ProjectToTemplateFragment.kt
```diff
@@ -0,0 +1,227 @@
...
+            val current = logText?.text
+            logText?.text = if (current.isNullOrEmpty()) line else "$current\n$line"
+        }
+    }
+
+    private fun showStatus(message: String, isError: Boolean) {
+        statusText?.apply {
+            text = message
+            setTextColor(
+                if (isError) Color.parseColor("#B3261E") else Color.parseColor("#0D6B42")
```
jatezzz: **Dark-mode status colors are hardcoded — the `values-night` variants are dead.** (should fix)

`values/colors.xml` and `values-night/colors.xml` both define `status_error_text` / `status_success_text`, but nothing references them — this line hardcodes the light-mode hex instead. So dark mode gets the light-mode success green (#0D6B42) on the IDE's dark surface, roughly 2:1 contrast.

```kotlin
setTextColor(ContextCompat.getColor(requireContext(),
    if (isError) R.color.status_error_text else R.color.status_success_text))
```

That also stops the two night colors from being dead resources.

project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/Templatizer.kt
```diff
@@ -0,0 +1,758 @@
...
+    val templateJsonFile = File(templateDir, "template.json")
+    templateJsonFile.writeText(buildTemplateJson(templateName, includeLanguage).toString(4) + "\n")
+    report.ok("$templateJsonFile")
+
+    val thumbPng = File(templateDir, "thumb.png")
+    thumbPng.writeBytes(Base64.decode(THUMB_PNG_B64, Base64.DEFAULT))
+    report.ok("$thumbPng (generic thumbnail, replace with an app-specific one if desired)")
+
+    val cgtFile = File(outputDir.parentFile, "$safeName.cgt")
+    zipDirectory(outputDir, cgtFile)
```
jatezzz, line 756: **Re-running with a different template name zips the previous run into the new `.cgt`.** (should fix)

`dest` (`outputDir/safeName`) is deleted if it exists, but `outputDir` itself is reused across runs, and `templates.json` is rewritten to list only the new path. Then this line zips *everything* under `outputDir`.

Convert as "Foo", then convert as "Bar" -> `Bar.cgt` contains the entire tokenized Foo tree alongside Bar. Either wipe/recreate `outputDir` at the start of a real run, or zip only `templates.json` + `safeName/` rather than the whole directory.

project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/Templatizer.kt
```diff
@@ -0,0 +1,758 @@
...
+    val languages = detectSourceLanguages(projectDir, module)
+    val isMixedLanguage = languages.size > 1
+
+    val basePackage = processAppBuildGradle(projectDir, module, report, dryRun, isMixedLanguage)
+
+    processStringsXml(projectDir, module, report, dryRun)
+    processJavaKtSources(projectDir, module, basePackage, report, dryRun)
+    flattenPackageDirectory(projectDir, module, basePackage, report, dryRun)
+
+    if (!skipCleanup) {
```
jatezzz, line 542: **`skipCleanup` suppresses the personal-info warning too — including keystores.** (should fix)

This gates `cleanupKeystores` *and* `flagPersonalInfoFiles` behind the same flag. Check "Skip build/ and keystore cleanup" and your `*.jks` / `*.keystore` / `*.p12` get copied into the bundle and zipped into the `.cgt` with no `[REVIEW]` line at all — the one run where the user most needs the warning is the one run that suppresses it.

Suggest: always run `flagPersonalInfoFiles`, and when cleanup is skipped, *flag* keystores instead of deleting them.

project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/Templatizer.kt
```diff
@@ -0,0 +1,758 @@
...
+ * Turns an Android Studio project into a Code On The Go (.cgt) template
+ * bundle by substituting concrete values (versions, package name, app name,
+ * SDK levels, Java compatibility levels) with Pebble tokens (${{ TOKEN }}),
+ * then writing templates.json and template/template.json alongside a
+ * thumbnail. Pure file-I/O logic with no dependency on the plugin API, so it
+ * can be unit-tested and called directly from the plugin's UI layer.
+ */
+
+/** Directory/file names skipped when copying the source project into the bundle. */
+private val COPY_IGNORE = setOf(
```
jatezzz, line 20: **`local.properties` ships in the `.cgt` by default.** (should fix)

`COPY_IGNORE` covers `.git` / `.gradle` / `.idea` / `release.properties` but not `local.properties`, which is only flagged by `flagPersonalInfoFiles` and still ends up zipped into the bundle.

Since the whole point of the output is distribution, `local.properties` should be in `COPY_IGNORE` outright — it's always machine-specific (absolute `sdk.dir`) and always regenerated on the consuming side. Leaving `google-services.json` as flag-only seems right; that one may be intentional.

project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/Templatizer.kt
```diff
@@ -0,0 +1,758 @@
...
+    while (f != null && f != root) {
+        if (f.name in COPY_IGNORE) return true
+        f = f.parentFile
+    }
+    return false
+}
+
+private fun copyProject(source: File, dest: File) {
+    source.walkTopDown()
+        .onEnter { it == source || it.name !in COPY_IGNORE }
```
jatezzz: **The entire `build/` tree is copied, then deleted.**

`COPY_IGNORE` skips `.gradle` but not `build`, and `cleanupBuildDirs` removes it from the copy afterward. On a real Android project that's hundreds of MB of duplicated file IO on a phone before anything useful happens.

Adding `"build"` to the copy-time skip when `!skipCleanup` gets the same result for a fraction of the IO.

project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/fragments/ProjectToTemplateFragment.kt
```diff
@@ -0,0 +1,227 @@
...
+                if (installed) "[OK]      Installed template: $cgtFile"
+                else "[SKIP]    Could not install template: $cgtFile"
+            )
+        }
+    }
+
+    private fun appendLogLine(line: String) {
+        activity?.runOnUiThread {
+            val current = logText?.text
+            logText?.text = if (current.isNullOrEmpty()) line else "$current\n$line"
```
jatezzz: **Log rendering is O(n^2) on the main thread.**

This rebuilds the whole log string per line, with one `runOnUiThread` hop per line. Invisible on the 15-file project in the PR description, quadratic on a large one.

`logText?.append("$line\n")` fixes it, or batch the lines before posting.

project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/fragments/ProjectToTemplateFragment.kt
```diff
@@ -0,0 +1,227 @@
...
+                showStatus("Conversion failed: ${e.message}", isError = true)
+            } finally {
+                setRunning(false)
+            }
+        }
+    }
+
+    private fun onInstallClicked() {
+        val cgtFile = pendingCgtFile ?: return
+        installButton?.isEnabled = false
```
jatezzz, line 171: **A failed install can't be retried.**

The button is disabled here and nothing re-enables it, so if `registerTemplate` fails (or the service lookup returns null) the user has to re-run the whole conversion to try again. A `finally { installButton?.isEnabled = true }` around the launch body covers it.

project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/Templatizer.kt
```diff
@@ -0,0 +1,758 @@
...
+import java.util.zip.ZipEntry
+import java.util.zip.ZipOutputStream
+
+/**
+ * Turns an Android Studio project into a Code On The Go (.cgt) template
+ * bundle by substituting concrete values (versions, package name, app name,
+ * SDK levels, Java compatibility levels) with Pebble tokens (${{ TOKEN }}),
+ * then writing templates.json and template/template.json alongside a
+ * thumbnail. Pure file-I/O logic with no dependency on the plugin API, so it
+ * can be unit-tested and called directly from the plugin's UI layer.
```
jatezzz: **The "can be unit-tested" claim is contradicted by the code and the build config.**

The file is plugin-API-free, but it uses `android.util.Base64` (`THUMB_PNG_B64` decode) and `org.json`. With `unitTests.isReturnDefaultValues = true` in `build.gradle.kts`, `Base64.decode` returns `null` in exactly the JVM test this comment promises -> NPE in `writeBytes`. There are no tests in the PR.

`minSdk = 26`, so `java.util.Base64` works and is JVM-testable. Worth switching and adding a test for `sanitizeTemplateName` plus one substitution path, or dropping the claim.

project-to-template/src/main/res/layout/fragment_main.xml
```diff
@@ -0,0 +1,113 @@
...
+            android:layout_marginTop="12dp"
+            android:gravity="center_vertical"
+            android:orientation="horizontal">
+
+            <com.google.android.material.button.MaterialButton
+                android:id="@+id/convertButton"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:enabled="false"
+                android:text="Convert to Template" />
```
jatezzz: **UI strings are hardcoded in the layout and in the Fragment.**

`strings.xml` has only `app_name`, while "Convert to Template", "Dry run (preview only, writes nothing)", "Log", "Install Template", "Project", and (in the Fragment) "No project open" / "Conversion failed: ..." are all literals. Repo convention puts user-facing UI text in `strings.xml` — model-facing / protocol strings are the ones that stay inline.

project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/Templatizer.kt
```diff
@@ -0,0 +1,758 @@
...
+            """(id\s+"com\.android\.(?:application|library)"\s+apply\s+false\s+version\s+")([^"]+)(")""",
+        ),
+        "AGP_VERSION",
+        spaceAfterSuffix = true,
+    )
+    if (n == 0) {
+        report.skip("$path: no 'apply false version' plugin lines found, no changes made")
+        return
+    }
+    report.ok("$path: replaced $n AGP version occurrence(s)")
```
jatezzz: **Duplicate log line.** `report.ok(...)` here and then `writePeb` reports `ok` again for the same file, so the log shows two entries per processed file. Same pattern at line 336 in `processAppBuildGradle`.

project-to-template/src/main/kotlin/org/appdevforall/projecttotemplate/Templatizer.kt
```diff
@@ -0,0 +1,758 @@
...
+    if (!skipCleanup) {
+        cleanupBuildDirs(projectDir, report, dryRun)
+        cleanupKeystores(projectDir, report, dryRun)
+        flagPersonalInfoFiles(projectDir, report)
+    }
+
+    return languages
+}
+
+private fun isIgnoredUnderRoot(file: File, root: File): Boolean {
```
jatezzz: **`isIgnoredUnderRoot` is mostly redundant.** The `onEnter` prune in `copyProject` already stops descent into ignored directories (and the pruned directory isn't yielded), so the ancestor walk only ever matters for ignored *files* like `release.properties` — which a plain `it.name !in COPY_IGNORE` filter covers.

project-to-template/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+    id("com.itsaky.androidide.plugins.build")
+}
+
+pluginBuilder {
+    pluginName = "project-to-template"
+}
+
+android {
+    namespace = "org.appdevforall.projecttotemplate"
+    compileSdk = 34
```
jatezzz, line 15: **Merge-order note:** `compileSdk` / `targetSdk = 34` here will conflict with #54 (plugin toolchain standardization). Whichever lands second needs a rebase.

jatezzz (commented):
## Review

Reviewed against freshly-fetched `origin/main`. I did not build or device-test — taking the PR's "builds clean + emulator-verified end-to-end" at face value.

### What's right

The repo conventions are all here, including the ones that usually get missed: tooltip category is the full `"plugin_org.appdevforall.projecttotemplate"` form (not a slug), `TOOLTIP_TAG` is a shared const wired to the `NavigationItem`, Tier 3 `getTier3DocsAssetPath()` + a real `assets/docs/index.html` (white bg / black text), repo-root `../libs/*.jar` + `../gradlew` with no bundled copies, `Theme.Material3.DayNight` + `values-night/`, AAR metadata checks disabled, permission strings are exact enum values, `sanitizeTemplateName` genuinely kills traversal, views nulled in `onDestroyView`, dry run against a disposable temp copy, and both README + `update-libs.yml` MAP wiring are present.

### Summary of findings

Should fix before merge:

1. **Dark-mode status colors are hardcoded** — the `values-night` variants are dead resources.
2. **Re-running with a different template name zips the previous run into the new `.cgt`.**
3. **`skipCleanup` suppresses the personal-info warning too**, including keystores.
4. **`local.properties` ships in the `.cgt` by default.**

Worth fixing (device-level cost):

5. The entire `build/` tree is copied, then deleted.
6. Log rendering is O(n^2) on the main thread.

Minor: install button can't be retried after a failed install; the "unit-testable" KDoc claim is contradicted by `android.util.Base64` + `unitTests.isReturnDefaultValues = true`; UI strings hardcoded rather than in `strings.xml`; duplicate log lines; `isIgnoredUnderRoot` is redundant with the `onEnter` prune; `compileSdk`/`targetSdk = 34` will collide with #54.

Details are in the inline comments.

### Verdict

Approve once #1 and #2 are fixed — those two produce visibly wrong output. #3/#4 are judgment calls, but both cut against the plugin's purpose (shipping a template someone else installs), so worth folding in here. Everything else is follow-up material.

hal-eisen-adfa:
Thanks for the thorough review, @jatezzz — addressed everything in 2957e4e.

**Should-fix**
1. **Dark-mode colors** — `showStatus` now uses `ContextCompat.getColor(…, R.color.status_error_text / status_success_text)`, so `values-night/` is live.
2. **Re-run bundling** — a real run now wipes the whole `outputDir` (not just `dest/`) before copying, so a prior name's tree can't be zipped in. `.cgt` files live in `outputDir.parentFile`, so previous bundles are preserved.
3. **`skipCleanup` suppressing warnings** — `flagPersonalInfoFiles` now always runs; when cleanup is skipped, keystores are **flagged** (`[REVIEW] … remove it before distributing`) instead of silently shipped.
4. **`local.properties`** — added to `COPY_IGNORE` outright; `google-services.json` left flag-only as you suggested.

**Worth-fixing**
5. **`build/`** — added to the copy-time skip when `!skipCleanup`, so it's never copied (was copy-then-delete).
6. **O(n²) log** — switched to `logText.append("$line\n")`.

**Minor**
- Install button re-enables after a failed install (retry without re-converting).
- **`android.util.Base64` → `java.util.Base64`** (minSdk 26), and added JVM unit tests: `sanitizeTemplateName` (incl. traversal) + a dry-run substitution path. The dry-run pipeline touches no Android APIs, so it's safe under `isReturnDefaultValues`.
- UI strings moved to `strings.xml`.
- Removed the duplicate per-file `[OK]` lines (`writePeb` already logs) at the three sites.
- Deleted `isIgnoredUnderRoot` — the `onEnter` prune + `name !in ignore` filter cover it.

**Note (12)** — left `compileSdk`/`targetSdk = 34` as-is. #54 (SDK 36) is still open and owns the toolchain bump; whichever lands second rebases, as you noted.

**Verification** — `testDebugUnitTest` + `assemblePlugin` green. Beyond the unit tests I ran the *real* (non-dry) `createTemplateBundle` twice via a JVM harness (`Foo` then `Bar`) and asserted the resulting `.cgt` contains no stale `Foo/` subtree, no `local.properties`, no `build/`, and no `.jks` by default — and that `skipCleanup` retains **and** flags the keystore. All passed.

hal-eisen-adfa:
### Device verification (emulator, Code On The Go)

Installed the debug `.cgp` via **IDE preferences → Plugin Manager** (uninstalled the prior build, installed the new one, restarted) and drove the full flow against a real "My Application" project that contains a `release.keystore`, `release.properties`, and a `build/` dir.

- **#2 (re-run isolation)** — Converted as **Alpha**, then as **Beta**. `Beta_Template.cgt` (pulled + `unzip`) contains only `Beta_Template/` + `templates.json` — **no stale `Alpha_Template/` subtree**. The log's first line on the second run is `[REMOVED] …/My Application-cgt` (the new output-dir wipe). Both `Alpha_Template.cgt` and `Beta_Template.cgt` are preserved (the wipe only touches the intermediate bundle dir).
- **#5 (build/) + #3 (keystore) default run** — `Alpha`/`Beta` `.cgt` contain **no `build/`, no `release.keystore`, no `.gradle`/`.cg`/`.idea`**, even though the source project still has them. Summary: "removed 1" (the keystore).
- **#3 (skip cleanup)** — With "Skip build/ and keystore cleanup" checked (**Gamma**), the `.cgt` **retains** `release.keystore` and `build/`, and the log shows the new flag line: `[REVIEW] …/release.keystore is a signing keystore; remove it before distributing this template` (summary "1 to review"). Old code shipped it silently.
- **#1 (dark-mode colors)** — Toggled the device to dark mode and re-ran. The success status renders as the `values-night` mint green (`#8CD9B0`), clearly readable on the dark surface — not the old hardcoded `#0D6B42`.
- **#7 (install)** — "Install Template" registered via `IdeTemplateService` (`[OK] Installed template: …`); button stays disabled after a successful install (the `if (!installed)` re-enable is the retry-on-failure path).
- **#8/#10** — `template/thumb.png` decodes fine (`java.util.Base64`), and each processed file logs exactly one `[OK]` line (no duplicates).

`.cgt` contents confirmed by pulling each bundle off the device and inspecting with `unzip`.

jatezzz (approved):

PR #63, opened 2026-08-06
ADFA-1798 | Add proactive memory warning for custom model loading

jatezzz:
## Description

This PR introduces a "pre-flight" memory check to prevent application crashes caused by loading oversized custom `.gguf` models into memory. Before attempting to load a selected model, the application now reads the file's GGUF header to estimate its required footprint (weights plus KV cache and compute buffers) and compares it against the device's currently available RAM.

If the model is deemed too large and poses a crash risk, a non-blocking warning dialog alerts the user. The dialog provides a clear explanation of the memory shortfall and offers two choices: "Cancel" (aborting the load safely, which is the default) or "Proceed anyway" (empowering power users to attempt the load if they plan to free up RAM). This proactively enhances app stability and builds user trust.

## Details

* **Memory Estimation:** Added `GgufHeaderReader` to parse metadata blocks and `ModelMemoryEstimator` to calculate load and runtime byte requirements.
* **System RAM Check:** Added `DeviceMemory` interface and `SystemDeviceMemory` implementation to read real-time available RAM from `ActivityManager`.
* **Decision Logic:** Introduced `ModelMemoryGate` to evaluate estimates against available RAM, assigning a risk severity of `TIGHT` or `INSUFFICIENT`.
* **User Interface:** Implemented `MemoryWarningDialogFragment` to present the warning, utilizing a new `UserConfirmation` coroutine primitive to gracefully handle async user decisions and process death.
* **Documentation:** Updated `index.html` and tooltips to explain the memory requirements and warning behaviors to users.

https://github.com/user-attachments/assets/fe9028bc-88c8-4d02-8f02-6bd5d2c85831

## Ticket

[ADFA-1798](https://appdevforall.atlassian.net/browse/ADFA-1798)

## Observation

* The feature is designed to fail open: if the model's file size or the device's free RAM cannot be read, or if the GGUF header is unparseable, the app defaults to loading the model without a warning to avoid blocking legitimate user flows.
* All file reading and `ContentResolver` queries during the pre-flight check are dispatched to the IO thread to avoid blocking the main UI thread.

[ADFA-1798]: https://appdevforall.atlassian.net/browse/ADFA-1798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ai-assistant/src/main/assets/docs/index.html
+++ b/ai-assistant/src/main/assets/docs/index.html
@@ -56,6 +56,28 @@
       agent reads are sent to Google over HTTPS.</li>
   </ul>
 
+  <h2>Will this model fit in memory?</h2>
+  <p>When you pick a local <code>.gguf</code> file, it is measured against the RAM
+  free on the device at that moment, before anything is loaded. If it looks too
+  large you get a warning with the actual figures and two choices:</p>
+  <ul>
+    <li><b>Cancel</b> — nothing is saved and nothing is loaded. Whatever model you
+      had selected before stays in use. Dismissing the warning with Back does the
+      same thing.</li>
+    <li><b>Proceed anyway</b> — the model is accepted despite the shortfall. Use
+      this when you are about to close other apps, for example.</li>
+  </ul>
+  <p>The warning quotes two numbers. Memory <b>to load</b> is the weights: these are
+  memory-mapped, so they need not all fit at once — when they don't, the device pages
+  them in and out, which is why an oversized model can stall for minutes instead of
+  failing immediately. Memory <b>to run</b> is the KV cache and compute buffers, which
+  are ordinary allocations and do have to fit. That is why a model can be reported as
+  risky rather than impossible: the outcome genuinely depends on how much paging your
+  device will tolerate.</p>
+  <p>To fit a large model, close other apps and select it again, or pick a smaller or
+  more heavily quantized build. A Q4_K_M quantization of a 1–3B model is the most
+  likely to run comfortably.</p>
+
   <h2>What the agent can do</h2>
   <ul>
     <li>Read and search files <b>within the current project</b>.</li>
--- a/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/AiAssistantPlugin.kt
+++ b/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/AiAssistantPlugin.kt
@@ -66,6 +66,10 @@ class AiAssistantPlugin : IPlugin, UIExtension, DocumentationExtension, Settings
         const val TOOLTIP_TAG_SETTINGS_GEMINI_MODEL = "ai_settings_gemini_model"
         const val TOOLTIP_TAG_SETTINGS_GET_KEY = "ai_settings_get_free_key"
 
+        // Tags for the memory pre-flight warning (see MemoryWarningDialogFragment).
+        const val TOOLTIP_TAG_MEMORY_PROCEED = "agent_memory_warning_proceed"
+        const val TOOLTIP_TAG_MEMORY_CANCEL = "agent_memory_warning_cancel"
+
         @Volatile
         private var pluginContext: PluginContext? = null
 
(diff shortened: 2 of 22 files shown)


hal-eisen-adfa:
### Code review

Found 1 issue:

1. Unbounded `Long` multiplication of header-supplied values overflows silently, producing a wrong memory estimate (bug due to `kvCacheBytes` multiplying five values that are only floor-checked with `takeIf { it > 0L }`). `GgufHeaderReader` bounds entry counts, array lengths, and string lengths, but never the value of a scalar shape field, so a corrupt or crafted `.gguf` can wrap the product. Two outcomes, both executed against this branch:

   - `block_count = 2^61`, `key_length = value_length = 288` — the KV term wraps to exactly `0`, so `runBytes` collapses to the flat 256 MB compute-buffer allowance. The estimate is understated and the severity is downgraded (`TIGHT` where the real shape would be `INSUFFICIENT`).
   - `block_count = 2^49`, `key_length = value_length = 1` — the product is exactly `2^63`, i.e. `Long.MIN_VALUE`, so `totalBytes` goes negative (`-9223372032586340352`). `availableBytes >= estimate.totalBytes` then holds for any device and the gate returns `Verdict.Safe`, skipping the warning entirely.

   That is the "wrong estimate" the reader's own KDoc says must never happen — "fails closed to null: an unreadable header must mean 'no estimate', never a wrong one". A cheap fix is an upper bound on the shape fields (or `Math.multiplyHigh`/`Math.*Exact` with a catch) returning null on overflow, so the estimate stays absent rather than wrong. `GgufHeaderReaderTest` covers corrupt entry counts and string/array lengths, but no corrupt scalar shape value.

https://github.com/appdevforall/plugin-examples/blob/17a41730b9ac47badf98f0d738057c35b38c59c6/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/memory/ModelMemoryEstimator.kt#L79-L83

https://github.com/appdevforall/plugin-examples/blob/17a41730b9ac47badf98f0d738057c35b38c59c6/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/memory/ModelMemoryGate.kt#L46-L51

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

hal-eisen-adfa:
### Code review

The overflow finding from the previous review is fully addressed by the ceilings and the saturating `totalBytes` in `11dfc00`, including tests for both repro cases. Two new issues, both in the 5 lines that commit added to `ai-core`:

Found 2 issues:

1. The new pre-load gate hard-refuses loads that this PR's own "Proceed anyway" dialog just authorized. `ModelLoadDiagnostics.diagnose` was a post-mortem attributor — on `main` it is called only from the `catch` around `llama.load()`, and its KDoc explains the `max(256 MB, size/4)` headroom is deliberately conservative "because overestimating it would blame a corrupt model on memory". Promoting it to a gate makes it disagree with the ai-assistant estimate that drove the dialog: for a 8.5 GB Q8_0 with a parsed header (`runBytes` = 768 MB) on a device with 1.5 GB free, the gate returns `TIGHT`, the user proceeds, the path is persisted and settings shows "Loaded" — then the first message throws `ModelLoadException` quoting 2.1 GB, a number the user never saw, with no override and without `llama.load()` ever being attempted. `ModelLoadException` is not caught on the send path, so the model becomes permanently unusable. This contradicts `ModelMemoryGate`'s stated invariant.

https://github.com/appdevforall/plugin-examples/blob/11dfc00bf85f091abb7d26c63bd241a4bb7d665e/ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/LocalLlmBackend.kt#L248-L253

https://github.com/appdevforall/plugin-examples/blob/11dfc00bf85f091abb7d26c63bd241a4bb7d665e/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/memory/ModelMemoryGate.kt#L4-L8

2. The gate reads free RAM before the previous model is unloaded, so a switch is measured against memory the switch itself would release — and a refusal never unloads, so retries fail identically. The check sits between the embedding-model classification and the unload block. The classification is placed there for a documented reason (ADFA-4388: "Classify BEFORE unloading any working chat model"), but that reasoning holds because `classify` is a static property of the file; free RAM is not. `LLamaAndroid.unload()` frees the context, batch and sampler — anonymous, non-reclaimable allocations (roughly 1 GB for a 7B at the hardcoded context) that `ActivityManager.MemoryInfo.availMem` excludes while resident. Switching from a large resident model to another can therefore trip `LowMemory`, and since the throw returns before the unload block, `modelLoaded`/`currentModelPath` are untouched and no user-facing path reclaims that memory — only `close()` at plugin disposal unloads. Every retry reproduces the verdict. Pre-PR, the same switch unloaded first and had a real chance of succeeding.

https://github.com/appdevforall/plugin-examples/blob/11dfc00bf85f091abb7d26c63bd241a4bb7d665e/ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/LocalLlmBackend.kt#L248-L261

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

hal-eisen-adfa (approved):

PR #64, opened 2026-08-07
ADFA-3070 | Fix agent chat unscrollable in landscape

jatezzz:
## Description

Fixed an issue where the agent chat is not scrollable on shorter screens, such as in landscape orientation. The previous `LinearLayout` setup caused the `RecyclerView` to be measured at zero height once the UI chrome exceeded the container's height. The layout has been updated to bound the chrome elements and float the composer toggle over the list. Additionally, an auto-hide controller was introduced to fold the composer away on compact screens, ensuring maximum space for the chat history.

## Details

* Extracted the composer behavior into a dedicated `ComposerAutoHideController` to manage visibility state and auto-hide countdowns.
* Added integer resources to define compact screens (under 480dp) and trigger auto-hiding after 5 seconds of idle time.
* Implemented window inset handling to properly pad the `RecyclerView` around display cutouts in landscape orientation.
* Consolidated the attachment, input, and send controls into a single bordered composer box layout.

https://github.com/user-attachments/assets/2542ce74-6624-4591-bbd4-f8741b041427

## Ticket

[ADFA-3070](https://appdevforall.atlassian.net/browse/ADFA-3070)

## Observation

The auto-hide functionality includes explicit safeguards to ensure the composer remains visible if the user is typing, the input is not empty, the agent is actively running, or if touch exploration accessibility services are enabled.

[ADFA-3070]: https://appdevforall.atlassian.net/browse/ADFA-3070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/ChatFragment.kt
+++ b/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/ChatFragment.kt
@@ -1,17 +1,22 @@
 package com.itsaky.androidide.plugins.aiassistant.fragments
 
+import android.content.res.Configuration
+import android.graphics.Rect
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.doOnAttach
 import androidx.core.view.isVisible
-import androidx.core.widget.doAfterTextChanged
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.lifecycleScope
 import androidx.lifecycle.repeatOnLifecycle
 import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
 import com.google.android.material.chip.Chip
 import com.google.android.material.snackbar.Snackbar
 import com.itsaky.androidide.plugins.PluginContext
@@ -21,6 +26,7 @@ import com.itsaky.androidide.plugins.aiassistant.R
 import com.itsaky.androidide.plugins.aiassistant.adapters.ChatAdapter
 import com.itsaky.androidide.plugins.aiassistant.databinding.FragmentChatBinding
 import com.itsaky.androidide.plugins.aiassistant.models.AgentState
+import com.itsaky.androidide.plugins.aiassistant.models.isRunning
 import com.itsaky.androidide.plugins.aiassistant.viewmodel.ChatViewModel
 import com.itsaky.androidide.plugins.base.PluginFragmentHelper
 import com.itsaky.androidide.plugins.services.IdeProjectService
@@ -48,6 +54,11 @@ class ChatFragment : Fragment(), ApprovalDialogFragment.Host {
     private lateinit var chatAdapter: ChatAdapter
     private lateinit var markwon: Markwon
     private val contextFiles = mutableListOf<File>()
+
+    private var composer: ComposerAutoHideController? = null
+
+    /** The message list's layout-declared padding, before any cutout inset is added. */
+    private val basePadding = Rect()
 
     private val tooltipService: IdeTooltipService? by lazy {
         try {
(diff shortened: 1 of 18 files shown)


hal-eisen-adfa:
### Code review

Found 1 issue:

1. The composer can reach an unrecoverable hidden state after process death. `attach()` passes the saved `visible` flag straight through, and `applyCompactRule` re-derives only `autoHide` — not `visible`. Sequence: fold the composer on a compact screen, the process is killed while backgrounded, the app restores into a taller configuration. `autoHide` is correctly recomputed to `false`, but the stale saved `visible = false` is still applied, so `setVisible` hides the input bar *and* hides the tab that would restore it (`binding.btnShowComposer.isVisible = autoHide && !visible` → `false && true`). The user is left with no input bar, no attach button, and no on-screen affordance to recover the composer.

https://github.com/appdevforall/plugin-examples/blob/26995d619bde752024105b9ab11f688603e8444f/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/ComposerAutoHideController.kt#L64-L68

https://github.com/appdevforall/plugin-examples/blob/26995d619bde752024105b9ab11f688603e8444f/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/ComposerAutoHideController.kt#L148-L153

`onConfigurationChanged` already guards the live-rotation case with `applyCompactRule(config, visible = true)`, but `attach()` never got the equivalent guard:

https://github.com/appdevforall/plugin-examples/blob/26995d619bde752024105b9ab11f688603e8444f/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/ComposerAutoHideController.kt#L80-L84

The KDoc on `saveState` already names this exact scenario — "the process can come back in an orientation other than the one it died in, and the saved flag would describe a screen that is gone" — but only `autoHide` is re-derived on restore, not `visible`:

https://github.com/appdevforall/plugin-examples/blob/26995d619bde752024105b9ab11f688603e8444f/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/ComposerAutoHideController.kt#L126-L135

Suggested fix: in `applyCompactRule`, ignore the restored `visible = false` once `autoHide` resolves to `false`, so the composer can only stay folded on a screen that still offers the reopen tab.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

hal-eisen-adfa:
### Code review

Found 1 issue:

1. The manual fold path has no `agentRunning` guard, so the Hide tab can fold the Stop button away mid-run (bug due to `ComposerAutoHideController.kt`: `canAutoHide()`'s `return !agentRunning` is consulted only by `schedule()`, so it blocks the idle-timeout fold but not a tap on `btnHideComposer`)

The KDoc on `onAgentRunningChanged` states that "a running agent also pins the composer open for as long as the run lasts":

https://github.com/appdevforall/plugin-examples/blob/3793da99a7f7b7d27ff8b5c5dc4e064232f85a57/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/ComposerAutoHideController.kt#L83-L88

But `setVisible` never checks `agentRunning`:

https://github.com/appdevforall/plugin-examples/blob/3793da99a7f7b7d27ff8b5c5dc4e064232f85a57/ai-assistant/src/main/kotlin/com/itsaky/androidide/plugins/aiassistant/fragments/ComposerAutoHideController.kt#L151-L165

On a compact screen this is reachable on exactly the screen class the PR targets: `onAgentRunningChanged(true)` calls `setVisible(true)`, which makes `btnHideComposer` visible (`autoHide && visible`), and its click listener routes straight to `setVisible(false)`, hiding `inputBarCard` and the Stop button it carries.

Recovery is a single tap on the still-visible `btnShowComposer`, so this is low severity. Either honour `agentRunning` on the manual path, or soften the KDoc so it does not promise a guarantee the code does not make.

Reviewed against base `feat/ADFA-1798-model-memory-preflight` (stacked PR), so PR #63's commits were excluded. Code review only, not device-verified.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jatezzz:
@hal-eisen-adfa Good catch on the mismatch. Keeping the manual path as-is and fixing the doc instead: folding the composer to read streaming output on a short landscape screen is a reasonable thing to want mid-run, and gating the tap would leave a visible button that does nothing. Reworded the KDoc so it only promises what the code does, the run opens the composer and suppresses the idle countdown; an explicit Hide still folds it, with Stop one tap away on the reopen tab.

hal-eisen-adfa (approved):

PR #65, opened 2026-08-11
ADFA-5096 | Split the AI backends into standalone plugins

jatezzz:
## Description

Reshapes the AI plugin set from `ai-core` + `ai-assistant` into **one agent plugin plus one plugin per inference backend**, so a user installs only the engine they want and a new backend can ship without touching this repo.

- **`ai-core`** — the Agent chat (tool layer, view models, UI, all folded in from `ai-assistant`, which is deleted) plus the `LlmInferenceService` router. Ships no model; discovers installed engines through `BackendRegistry` and names no provider anywhere in its code.
- **`ai-agent-llm`** — on-device `.gguf` backend (bundled llama.cpp AAR, `native.code`), registers as `local`.
- **`ai-agent-gemini`** — Gemini API backend (`network.access`), registers as `gemini`.

Along the way: each backend owns its preferences instead of reading `ai-core`'s `AgentSettings`; each contributes its own settings pane, system prompt and tests; the ADFA-5095 `LlmBackend` split is adopted (`getConfigSpecs`/`settingsFragmentClassName` moved onto `ConfigurableBackend`, `supportsHistory` declared, backend closed on `deactivate()` so disabling no longer orphans a live instance); and every plugin's sources are laid out by layer (`plugin/`, `backend/`, `preferences/`, `services/`, `logging/`…) with `utils/` and `data/` gone and logs tagged per plugin.

Because plugins load in **parallel with no ordering guarantee**, settings had to stop depending on who activated first. Each backend now reads the host's legacy store directly and migrates once from every prior location (old plugin id from before the `ai-backend-*` → `ai-agent-*` rename, `ai-core`, `ai-assistant`, the host's `LlamaPrefs`), guarded by a `migrated_*` flag so a value changed since is never clobbered. A device configured on the old build stays configured.

## Details

7 commits, 233 files, mostly moves/renames:

| | |
|---|---|
| `ai-assistant/` | deleted — merged into `ai-core` |
| `ai-backend-local/` → `ai-agent-llm/` | folder, package, `plugin.id`, label (submodule path in `.gitmodules` follows) |
| `ai-backend-gemini/` → `ai-agent-gemini/` | folder, package, `plugin.id`, label |

Plugin ids after the rename: `…plugins.aicore`, `…plugins.aiagentllm`, `…plugins.aiagentgemini`. README's examples table rewritten to state the new dependency shape (AI Core is mandatory; install at least one backend).

Tests moved with the code they cover and now sit under each owning plugin — 31 test files across the three plugins, all passing.

https://github.com/user-attachments/assets/7566fc16-4c21-4ce7-8793-041ceeacd38d

## Ticket

[ADFA-5096](https://appdevforall.atlassian.net/browse/ADFA-5096)

[ADFA-5096]: https://appdevforall.atlassian.net/browse/ADFA-5096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hal-eisen-adfa:
This depends on https://github.com/appdevforall/CodeOnTheGo/pull/1660 landing first

hal-eisen-adfa (commented):
These were reviewed statically and have not been verified on a device.

Three comments land at file level rather than inline because their target lines fall outside the diff hunks (two files are renames, one target is unchanged context); each names its line in the first line of the comment.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/AiSettingsFragment.kt
```diff
@@ -0,0 +1,262 @@
...
+        )
+        adapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item)
+        backendSpinner.adapter = adapter
+
+        wireTooltip(backendSpinner, AiCorePlugin.TOOLTIP_TAG_SETTINGS_BACKEND)
+
+        // A stored selection whose backend was uninstalled falls back to the first one offered,
+        // rather than leaving the screen describing a backend that cannot run.
+        val storedId = BackendRegistry.selectedId()
+        val selected = backends.firstOrNull { it.id == storedId } ?: backends.first()
```
hal-eisen-adfa: **Opening Agent settings silently persists `gemini` as the backend.**

On a fresh install of ai-core + ai-agent-local + ai-agent-gemini, `BackendRegistry.selectedId()` returns `null`. `backends` comes from `BackendRegistry.options()`, which is `.sortedBy { it.displayName }` — `"Gemini API"` sorts before `"Local LLM"` — so `backends.first()` is Gemini. Lines 157-159 then see `selected.id != storedId` and call `BackendRegistry.select("gemini")`, writing it to disk.

The user never touched the spinner. This contradicts `AiBackend.DEFAULT_ID = "local"` (`backends/AiBackend.kt:25`), which both `effectiveBackendId()` and `ChatViewModel` still treat as the default — and it silently opts a user into a cloud backend.

Suggest either not writing on the fallback path, or preferring `AiBackend.DEFAULT_ID` before falling back to first-by-label.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -596,47 +572,43 @@ class ChatViewModel(
      * Retries with delays to handle plugin loading timing.
      */
     fun checkBackendAvailability() {
-        android.util.Log.d("ChatViewModel", "checkBackendAvailability: Starting check")
+        android.util.Log.d(TAG, "checkBackendAvailability: Starting check")
         viewModelScope.launch(Dispatchers.IO) {
```
hal-eisen-adfa: **`checkBackendAvailability()` races itself; a stale run can clobber a newer result.**

`ChatFragment.onResume()` calls this unconditionally, and each call starts a new 2.5-second retry coroutine without cancelling the previous one.

Repro: open the Agent tab with no key configured — run A begins and will write `false` at t≈2.5s (line 632). At t≈1s the user opens Agent settings, saves a Gemini key, and returns; `onResume` fires run B, which finds the backend on attempt 1 and writes `true`. Run A then writes `false` on top of it. `sendMessage()` (line 648) now rejects with "No LLM backend is set up yet" even though one is configured, until the next resume.

Both runs also write the non-`@Volatile` `currentBackendId` from IO threads while `sendMessage` reads it from Main (lines 668/688), so the config can name a different backend than the one just resolved.

Cancelling the previous job (or keeping a single job) before launching would fix both.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/backends/BackendRegistry.kt
```diff
@@ -0,0 +1,110 @@
...
+        return backends.mapNotNull(::describe).sortedBy { it.displayName }
+    }
+
+    /**
+     * The loader that can instantiate [fragmentClassName], found by asking each registered backend
+     * which pane it contributes. Used by [BackendFragmentFactory] to rebuild a restored pane.
+     *
+     * @return the owning backend plugin's loader, or null if no installed backend claims the class
+     */
+    fun classLoaderFor(fragmentClassName: String): ClassLoader? =
```
hal-eisen-adfa: **`classLoaderFor()` matches only a backend's exact `settingsFragmentClassName`, so no other fragment from that plugin survives rotation.**

`LocalLlmSettingsFragment` shows `MemoryWarningDialogFragment` through its own `childFragmentManager`, which inherits `AiSettingsFragment`'s `BackendFragmentFactory`.

Repro: select a large `.gguf` so the memory warning dialog appears, then rotate. `FragmentManager` calls `instantiate(hostClassLoader, "…aiagentlocal.settings.MemoryWarningDialogFragment")`; `resolve()` returns `null` because that name is not any backend's `settingsFragmentClassName`, so it falls through to `instantiateOwn` → `delegate.instantiate(hostClassLoader, …)` → `ClassNotFoundException` (the class lives in ai-agent-local's `DexClassLoader`) → swallowed → a plain `MissingBackendFragment`.

The dialog disappears, the pending `UserConfirmation` is never answered, and `dropStaleMemoryWarning()`'s `as? MemoryWarningDialogFragment` cast yields `null`, so the stray fragment is never dismissed.

The resolve hook probably wants to be keyed on which backend plugin's loader can see the class (package or loader lookup) rather than on settings-pane-name equality.

Minor, same function: `options()` walks every registered backend and calls `describe()` on each (two cross-classloader accessor calls plus a `sortedBy`) just to find one class name, and `FragmentManager` calls the factory once per fragment instantiated *and* per fragment restored. A single id→option map built once per screen would avoid the repeated walk. Line 3 also imports `AiBackend` from inside its own package.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/plugin/GeminiPlugin.kt
```diff
@@ -0,0 +1,275 @@
...
+ * Registers the Google Gemini API backend with AI Core's inference router.
+ *
+ * Owns the transport *and* the UI that configures it: the backend names a settings Fragment that
+ * ships in this plugin, which whichever screen offers a backend selector mounts under its own
+ * selector. AI Core owns routing; nothing outside this plugin handles the API key.
+ */
+class GeminiPlugin : IPlugin, DocumentationExtension {
+
+    private lateinit var context: PluginContext
+    private var backend: GeminiBackend? = null
```
hal-eisen-adfa: **`backend` is written and read across threads without a happens-before edge, so registration can be silently skipped.**

`registered` is correctly `@Volatile` (line 27), but `backend` is a plain field. `activate()` writes it on the loading thread and then adds the listener. Because plugins load in parallel — this PR's own premise — the host may deliver `onPluginActivated("…aicore")` on a different thread; `registerBackend()` then reads `backend ?: return false` (line 138) with no synchronisation and can observe `null`.

When that happens `registered` stays `false`, no retry is scheduled, and the backend never appears in the registry — the user sees an empty backend selector until the IDE is restarted.

`LocalLlmPlugin.kt:24` has the identical shape. Marking `backend` `@Volatile` (or publishing it through the same volatile write as `registered`) would close it.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/AiSettingsFragment.kt
```diff
@@ -0,0 +1,262 @@
...
+
+    /** Backend whose pane is currently mounted, so re-selecting it does not rebuild it. */
+    private var shownBackendId: String? = null
+
+    companion object {
+        /** Tag for the mounted backend pane, so it can be found across a configuration change. */
+        private const val TAG_BACKEND_PANE = "backend_settings_pane"
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
```
hal-eisen-adfa, line 56: **The Material-transition guard was dropped in the move to ai-core.**

`origin/main`'s `ai-assistant/…/fragments/AiSettingsFragment.kt` set both in `onCreate()`:

```kotlin
enterTransition = null
exitTransition = null
```

with the comment *"Disable Material transitions to avoid resource loading issues — Plugin uses compileOnly dependencies, so Material transition resources aren't bundled"*. Grepping `enterTransition|exitTransition` across ai-core, ai-agent-local and ai-agent-gemini now returns nothing.

The screen is still mounted in the host's `PluginScreenActivity`, and this PR adds *more* fragment transactions than before (the backend pane replace/remove in `showBackendPane`/`clearPane`), so a default Material transition resolving against unbundled resources would now be hit on every backend switch, not only on screen entry.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsFragment.kt
```diff
@@ -0,0 +1,257 @@
...
+                if (!hasFocus) {
+                    viewModel.saveLocalModelSha256(text?.toString())
+                }
+            }
+        }
+        // On the labelled wrapper, not the field: long-press there is the paste menu.
+        view.findViewById<View>(R.id.local_model_sha_layout)
+            ?.let { wireTooltip(it, LocalLlmPlugin.TOOLTIP_TAG_SETTINGS_LOCAL_SHA) }
+
+        simplePromptCheckbox?.apply {
```
hal-eisen-adfa, line 154: **The "simple prompt" preference is write-only — and this PR adds a tooltip promising behaviour the code does not implement.**

`KEY_SIMPLE_PROMPT` / `use_simple_local_prompt` appears only in this checkbox (lines 144-150), the setter/getter (`LocalLlmSettingsViewModel:192-197`), the key declaration, and the `OWNED_KEYS` migration list. `LocalSystemPrompt.build(request)` takes only a `SystemPromptRequest` and never consults preferences, and `LocalLlmBackend` never reads the key — so toggling the checkbox changes the prompt not at all.

This PR compounds it by adding `LocalLlmPlugin.kt:248-256`, a new tooltip stating *"On by default. Small on-device models follow a short prompt far more reliably than the full tool-calling one; turn it off only if you are running a larger model"* — documentation for a code path that does not exist.

Either wire the flag into `LocalSystemPrompt.build`, or drop the control, its preference key, and the tooltip.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/plugin/LocalLlmPlugin.kt
```diff
@@ -0,0 +1,279 @@
...
+    /** True once [backend] is registered with the router, so re-registration is idempotent. */
+    @Volatile private var registered = false
+
+    companion object {
+        const val PLUGIN_ID = "com.itsaky.androidide.plugins.aiagentlocal"
+
+        /** Provider of [LlmInferenceService]; this plugin is useless without it. */
+        private const val AI_CORE_PLUGIN_ID = "com.itsaky.androidide.plugins.aicore"
+
+        private const val TOOLTIP_TAG_PLUGIN = "plugin_ai_backend_local"
```
hal-eisen-adfa: **These tooltip entries keep pre-rename tags and are registered without ever being reachable.**

`TOOLTIP_TAG_PLUGIN = "plugin_ai_backend_local"` here, and `"plugin_ai_backend_gemini"` in `GeminiPlugin.kt:35`, name the folders this PR renames away.

More importantly, neither constant is ever passed to `wireTooltip`, set as a `tooltipTag` on a `NavigationItem`/`TabItem`/menu item, or referenced from a manifest — these two plugins contribute no IDE UI of their own. The entries are written into the `Tooltips` table, but no long-press anywhere can reach them, which makes the Tier-3 `index.html` buttons (the only ones either plugin registers) unreachable from the IDE.

Per `CLAUDE.md`, a contributed element with no reachable tooltip fails review clause 6.7 — worth deciding whether these entries should hang off an ai-core-contributed element instead.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
hal-eisen-adfa, line 1: **(line 160 — outside the diff hunks, so this lands as a file-level comment.)**

**Four places disagree about the default backend, and none uses `AiBackend.DEFAULT_ID`.**

- `ChatViewModel.kt:160` hardcodes the literal `"local"`
- `checkBackendAvailability()` (lines 589-590) falls back to `backends.firstOrNull()?.id` — `ConcurrentHashMap` iteration order, i.e. registration order, which depends on which plugin activated first
- `selectedBackendLabel()` (~line 175) uses the same hash-order fallback
- `AiSettingsFragment.kt:156` uses `backends.first()` — alphabetical by display name

Meanwhile `AiBackend.DEFAULT_ID = "local"` (`backends/AiBackend.kt:25`) exists and is honoured only inside `effectiveBackendId`.

With both backends installed and nothing stored, the chat status line, the routed backend, and the settings pane can each name a different backend on the same launch — and the answer can change between launches. Routing all four through `AiBackend.DEFAULT_ID` would make this deterministic.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/backend/GeminiBackend.kt
hal-eisen-adfa, line 1: **(line 381 — outside the diff hunks, so this lands as a file-level comment.)**

**The no-arg `listModels()` omits the `scope.isActive` guard its sibling overload has.**

`listModels(apiKey)` at line 427 checks:

```kotlin
if (!scope.isActive) { future.completeExceptionally(...); return future }
```

with the comment *"close() cancels the scope, making launch a silent no-op; fail loudly instead"*. `listModels()` at lines 381-406 has no such check, so on a cancelled scope `scope.launch` is a no-op and nothing ever completes `future`. `BackendGeminiCatalogGateway.await()` then blocks on `future.get(60, SECONDS)` (`GeminiCatalogGateway.kt:96`).

Repro: open Agent settings → Gemini pane, disable the AI Agent Gemini plugin from the Plugin Manager (`GeminiPlugin.deactivate` → `releaseBackend` → `close()`), then tap Refresh Models. The button sits at "Loading" for a full 60s before falling back.

The same guard as line 427 would fix it.

ai-core/proguard-rules.pro
hal-eisen-adfa, line 1: **(line 4 — unchanged context outside the diff hunks, so this lands as a file-level comment.)**

**Both keep rules still name the pre-move class paths, and the by-name-instantiated settings fragment has none.**

- Line 4 keeps `com.itsaky.androidide.plugins.aicore.AiCorePlugin`; the class is now `…aicore.plugin.AiCorePlugin`
- Line 9 keeps `com.itsaky.androidide.plugins.aicore.LlmInferenceServiceImpl`; it is now `…aicore.services.LlmInferenceServiceImpl`

Both rules therefore match nothing. This PR edits the file (it drops the `LocalLlmBackend` and `com.itsaky.llama.**` rules) but misses the two renames.

Latent while `ai-core/build.gradle.kts:31` has `isMinifyEnabled = false` — but the moment minification is enabled, the manifest's `plugin.main_class` reflective load fails and the plugin will not start.

Separately, `AiSettingsFragment` is instantiated by name via `fragmentClassName` (`AiCorePlugin.kt:165`) and has no keep rule, whereas both sibling plugins keep theirs (`ai-agent-local/proguard-rules.pro:9`, `ai-agent-gemini/proguard-rules.pro:16`).

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -594,49 +578,56 @@ class ChatViewModel(
...
+                        android.util.Log.d(TAG, "checkBackendAvailability: Found ${backends.size} backends")
+
+                        // The selection is stored as the backend's own id, so it needs no mapping:
+                        // a backend this plugin has never heard of resolves like any other. With
+                        // nothing stored this must not fall to registration order, which is a hash
+                        // map's and so differs from what the settings screen shows.
+                        val preferredBackendId = AiBackend.preferredId(
+                            BackendRegistry.selectedId(),
+                            backends.map { it.id },
+                        )
```
hal-eisen-adfa: **F3 — the status label and the routing decision can name different backends.**

The comment here says the resolution "must not fall to registration order, which is a hash map's and so differs from what the settings screen shows" — but that is exactly what this call site does. The two paths feed `AiBackend.preferredId` differently ordered collections:

| Call site | `installedIds` source | Order |
|---|---|---|
| `ChatViewModel.kt:180`, `AiSettingsFragment.kt:162` (via `BackendRegistry.preferred`) | `BackendRegistry.options()` | `sortedBy { it.displayName }` |
| here (line 604) | `llmService.availableBackends` | `ConcurrentHashMap.values`, i.e. hash order |

`AiBackend.preferredId` only reaches its third branch — `installedIds.firstOrNull()` — when nothing is stored *and* no installed backend has id `local` (`DEFAULT_ID`). In that state the two orderings give different answers: e.g. with `ai-agent-gemini` plus some third-party `openai` backend installed, the settings screen and status line show whichever sorts first by display name, while the message is routed to whichever the hash map yields first.

Suggest routing this through `BackendRegistry.preferred(BackendRegistry.options())` like the other two call sites, so there is one ordering by construction rather than by convention. As written, the invariant the KDoc on `AiBackend.preferredId` states ("They must agree, or one launch can route to one backend while the label names another") is not enforced anywhere.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/AiSettingsFragment.kt
```diff
@@ -0,0 +1,269 @@
...
+        // Read once for the whole screen: the factory below is consulted for every fragment the
+        // panes create or restore, and each lookup would otherwise re-walk the registry.
+        backends = BackendRegistry.options()
```
hal-eisen-adfa, line 59: **F4 — a backend that activates after this screen opens is invisible until the screen is recreated.**

`backends` is read once here and never refreshed: there is no `onResume` re-read and no retry. The premise stated throughout this PR is that plugins load in parallel with no ordering — so opening Preferences → Configuration → Agent before `ai-agent-local`/`ai-agent-gemini` have called `registerBackend` leaves `backends` empty, the spinner hidden, and `R.string.backend_none_installed` on screen permanently. Backing out and re-entering is the only recovery, and nothing on screen suggests that.

The rest of the codebase already treats this race as real: `ChatViewModel.checkBackendAvailability()` retries 5 × 500 ms for precisely this, and the backend plugins install a `PluginLifecycleListener`. This screen has neither. Re-reading the registry in `onResume` (and rebuilding the selector when the set changed) would close it.

The caching rationale in the comment still holds for the `BackendFragmentFactory` lookup path — the point is only that the snapshot needs a refresh trigger, not that it should be re-walked per lookup.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/AiSettingsFragment.kt
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        val paneClass = loadPaneClass(backend, className)
+        if (paneClass == null) {
+            clearPane()
+            showPlaceholder(getString(R.string.backend_pane_unavailable))
+            return
+        }
+
+        backendSpecificContainer.removeAllViews()
```
hal-eisen-adfa: **F6 — `removeAllViews()` on a container the child FragmentManager owns, racing its own pending commit.**

Two problems with touching this container directly:

1. **Here it's redundant.** `replace(R.id.backend_specific_settings_container, …)` on the very next line already removes the previous pane's view. The extra `removeAllViews()` only defeats exit animations.
2. **In `showPlaceholder()` it's a race.** `clearPane()` posts `remove(pane)` — asynchronous — and then `showPlaceholder()` synchronously calls `removeAllViews()` on the same container and adds a `TextView`. When the FragmentManager later executes the pending transaction, it runs the outgoing pane's `onDestroyView` against a view already detached from its container, on a container whose only child is now the placeholder.

Suggest dropping the `removeAllViews()` calls and letting the FragmentManager own the container, with the placeholder as a sibling view toggled by visibility rather than a child of the fragment container.

hal-eisen-adfa (commented):
Review notes on the backend split. Ten findings, filed inline. Roughly ordered by how much they'd cost to hit:

- **F3** — the status label and the routing decision resolve the preferred backend from differently ordered lists, so they can disagree.
- **F4** — the settings screen reads the backend registry once in `onCreate` with no retry, so a late-activating backend is permanently invisible.
- **F6** — `removeAllViews()` races the child FragmentManager's pending commit.
- **F10** — three of the four `LlmInferenceServiceImpl` entry points hand the backend `config.backendId == "auto"`.
- **F11** — `CancellationException` is converted into a catalog failure, defeating the rethrow guard directly above it.
- **F7**, **F13** — two migrations that cannot fire on any real device.
- **F12** — ai-core's help text hardcodes the two first-party providers, contradicting the "ai-core names no provider" invariant.
- **F14**, **F15** — duplicated constants with hand-maintained sync contracts, plus stale KDoc and dead imports.

Everything below was checked against the source on `29a4224`; each comment says what it was verified against.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/services/LlmInferenceServiceImpl.kt
```diff
@@ -114,24 +123,28 @@ class LlmInferenceServiceImpl : LlmInferenceService {
...
-                override fun onError(error: String) = callback.onError(error)
-            }
-            if (backend is LocalLlmBackend) {
-                backend.generateStreamingWithHistory(history, prompt, config, streamCallback)
-            } else {
-                backend.generateStreaming(prompt, config, streamCallback)
-            }
-            return
+        // Route by what the backend declares it can do, degrading one capability at a time.
+        when (backend) {
```
hal-eisen-adfa, line 123: **F10 — only `generateCompletion` writes the resolved id back into the config.**

(Anchoring here because this file shows up as a rename with a partial diff; the observation covers the whole class.)

`generateCompletion` does `config.backendId = effectiveId` (line 62). `generateWithHistory` (line 102), `generateStreaming` (line 82) and `generateStreamingWithTools` — the dispatch right below — all resolve `effectiveId` and route correctly, but hand the backend a config whose `backendId` is still the `"auto"` sentinel.

So a backend that keys any behaviour off `config.backendId`, or a caller that reads it back to learn where the request actually went, sees `"auto"` for three of the four entry points and a real id for the fourth. Either set it in all four, or in none — the asymmetry is the part that will bite, since it makes the field's meaning depend on which overload you happened to call.

Simplest fix: move `config.backendId = effectiveId` next to the `effectiveBackendId(...)` call in each method, or factor the resolve-and-stamp into one private helper the four share.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiCatalogGateway.kt
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        return try {
+            CatalogResult.Success(future.get(LIST_MODELS_TIMEOUT_SECONDS, TimeUnit.SECONDS).orEmpty())
+        } catch (e: ExecutionException) {
+            // The API failure the backend reported; its message carries the HTTP status.
+            CatalogResult.Failed(e.cause ?: e)
+        } catch (e: CancellationException) {
+            logger?.warn("$TAG: listModels was cancelled by the backend", e)
+            CatalogResult.Failed(e)
```
hal-eisen-adfa: **F11 — this swallows coroutine cancellation and reports it as a catalog failure.**

On the JVM `kotlin.coroutines.cancellation.CancellationException` is a typealias for `java.util.concurrent.CancellationException` — the same class. So this `catch` fires for coroutine cancellation too, not just a backend-cancelled future.

Concretely: the user taps Save, `GeminiSettingsViewModel.verifyGeminiKey` launches on the IO dispatcher and blocks in `future.get(...)`. The pane is destroyed (backend selector switched, or a rotation), the scope cancels, the future is cancelled and `get()` throws. This line turns that into `CatalogResult.Failed`, whose `toKeyVerification()` yields Unreachable/Unknown — so `verifyGeminiKey` returns a *verdict* instead of propagating cancellation, and its own `catch (e: CancellationException) { throw e }` guard never sees it. The caller then runs `confirmSaveUnverified()` against a dead fragment.

Rethrow instead, and let the `ExecutionException` / `TimeoutException` branches keep handling the real failures:

```kotlin
} catch (e: CancellationException) {
    throw e
}
```

If a backend-initiated cancellation genuinely needs to surface as `Failed`, it has to be distinguished some other way than by exception type — the type can't tell the two apart.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/plugin/AiCorePlugin.kt
```diff
@@ -432,183 +436,122 @@ class AiAssistantPlugin : IPlugin, UIExtension, DocumentationExtension, Settings
...
+
+    /** Copies the old plugin's saved conversations into this plugin's storage. */
+    private fun adoptLegacyChatSessions() {
+        val legacyDir = File(
+            File(context.getAppFilesDir(), "plugins/$LEGACY_ASSISTANT_PLUGIN_ID"),
+            CHAT_SESSIONS_DIR
+        )
+        if (!legacyDir.isDirectory) return
+
+        val targetDir = File(context.getPluginFilesDir(), CHAT_SESSIONS_DIR)
```
hal-eisen-adfa: **F7 — this migration cannot fire on any device; the KDoc's claim is not accurate.**

Chat history isn't stored under `getPluginFilesDir()/chat_sessions`. `ChatStorageManager.kt:18` uses `context.getSharedPreferences("ai_assistant_chats", MODE_PRIVATE)` and writes the sessions JSON under the key `"chat_sessions"` (line 22/33) — an app-level prefs file, not a directory of files. Grepping `ai-core/src/main` shows `getPluginFilesDir()` referenced **only** inside `adoptLegacyChatSessions()` and `migrateChatHistory()`; nothing reads what they write.

So `legacyDir.isDirectory` is false on every device and this returns immediately. Conversations do survive the rename, but only incidentally — the prefs file name `ai_assistant_chats` is app-global and unchanged — which means the KDoc rationale ("Without this, upgrading looks exactly like a factory reset: … no conversations") credits this code with something it doesn't do.

Either delete both methods, or point them at the store that actually holds the data. Leaving them as-is is the worst option: a future reader will trust the comment.

Worth double-checking `migrateChatHistory()` (line 555) on the same basis — it copies `getAppFilesDir()/chat_sessions` → `getPluginFilesDir()/chat_sessions`, neither of which anything reads.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/preferences/GeminiPreferences.kt
```diff
@@ -0,0 +1,120 @@
...
+     * plugin's settings into its own file on activation, but this backend may migrate before that
+     * has happened, and would otherwise find the newer file empty and conclude there is nothing to
+     * take.
+     */
+    private val LEGACY_FILES = listOf(
+        OWN_LEGACY_FILE,
+        legacyFileName(LEGACY_PLUGIN_ID),
+        legacyFileName("com.itsaky.androidide.plugins.aicore"),
+        legacyFileName("com.itsaky.androidide.plugins.aiassistant"),
+    )
```
hal-eisen-adfa: **F13 — two of these legacy stores are keyed on plugin ids that never shipped.**

`git ls-tree -r origin/main` has no `ai-backend-gemini/` or `ai-backend-local/` folder — those ids (`…plugins.aigemini`, `…plugins.ailocal`) exist only in unmerged branch history (`260ed84`, `fdf65f9`, later renamed by `992bc8d`). No released `.cgp` ever wrote under them, so `plugin_…aigemini_GeminiSettings` and `plugin_…aigemini_AgentSettings` cannot hold anything on a user's device.

Two consequences:

- Every first run opens two extra `SharedPreferences` files (each a real file lookup) and reads them for nothing. `LocalLlmPreferences` does the same with `…ailocal` (lines 36, 53-54).
- The KDoc at 35-39 — "a device with a verified key would look unconfigured" — describes an upgrade path that never existed, which is the kind of comment that keeps dead code alive through the next three refactors.

`legacyFileName("…aicore")` and `legacyFileName("…aiassistant")` are the real ones and should stay. Suggest dropping `OWN_LEGACY_FILE` and `legacyFileName(LEGACY_PLUGIN_ID)` here and the matching pair in `LocalLlmPreferences`. If you'd rather keep them as cheap insurance for anyone running a dev build off those branches, that's defensible — but then the comment should say *that*, not imply a shipped release.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/plugin/AiCorePlugin.kt
```diff
@@ -432,183 +436,122 @@ class AiAssistantPlugin : IPlugin, UIExtension, DocumentationExtension, Settings
...
-                anyway.</p>
-                <p>Use the eye button to check what you typed, <b>Edit</b> to change
-                the key later and <b>Clear</b> to remove it from the device.</p>
-                <p>If the Keystore entry is ever lost — clearing the app's data,
-                for instance — the stored key can no longer be decrypted and must
-                be re-entered here.</p>
+                <p>Lists every AI backend plugin installed and registered with
+                <b>AI Core</b> — for example <b>AI Agent Local</b> for
+                on-device <code>.gguf</code> models, or <b>AI Agent Gemini</b>
+                for Google's cloud API.</p>
```
hal-eisen-adfa: **F12 — ai-core's help text names the two first-party providers, which is the thing the split was meant to stop.**

`BackendRegistry`'s KDoc states the invariant: "no provider is named anywhere in this plugin … one that ships in a `.cgp` written by someone else appears without a line of code changing here." The runtime code honours that. The user-facing strings don't:

- here — "for example **AI Agent Local** … or **AI Agent Gemini**"
- `TOOLTIP_TAG_TAB` — "Local — on-device inference via llama.cpp", "Gemini — Google's cloud API"
- `TOOLTIP_TAG_MESSAGE_OPEN_SETTINGS` — "no local model selected, or no Gemini API key saved"
- `res/values/strings.xml:157` (`backend_none_installed`) — names both again

So a user who installs only a third-party backend long-presses the selector and reads help about two backends they don't have, and shipping a new backend *does* mean editing ai-core after all.

The empty-list case can be phrased without naming anyone ("Install an AI backend plugin from the Plugin Manager"), and the per-backend description is something each backend already knows — it has a `displayName` and its own tooltip category. If naming the first-party pair is a deliberate onboarding call, that's a reasonable trade, but it should be written down as an exception rather than left contradicting the KDoc two files over.

ai-core/src/main/res/values/strings.xml
```diff
@@ -1,25 +1,162 @@
...
+    <!-- Settings: local model pane -->
+
+    <!-- Settings: models -->
+
+    <!-- Local model selection -->
+
+    <!-- Local model selection: memory pre-flight warning -->
+
+    <!-- Backend selector states; the panes themselves come from the backend plugins. -->
+    <string name="backend_none_installed">No AI backend is installed. Install a backend plugin, for example AI Agent Local for on-device models, or AI Agent Gemini for Google\'s API, from the Plugin Manager.</string>
```
hal-eisen-adfa: **F12 (cont.)** — same issue as the tooltip detail in `AiCorePlugin.kt:446`: this string names `AI Agent Local` and `AI Agent Gemini`, so ai-core can't gain a third-party backend without a string change here. A provider-neutral phrasing ("No AI backend is installed. Install a backend plugin from the Plugin Manager.") keeps the message useful and drops the coupling.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/GgufFileInspector.kt
```diff
@@ -1,23 +1,22 @@
...
+ * stream and never throw. Keep [GGUF_MAGIC] in step with `GgufModelInspector.GGUF_MAGIC` if the
+ * format ever changes.
  */
 internal object GgufFileInspector {
 
-    private const val TAG = "GgufFileInspector"
+    private const val TAG = "$LOG_PREFIX.GgufFileInspector"
 
     /** The GGUF magic: the four ASCII bytes a `.gguf` file starts with. */
     private val GGUF_MAGIC = "GGUF".toByteArray(Charsets.US_ASCII)
```
hal-eisen-adfa: **F14 — three definitions of the GGUF magic, now all in the same module.**

- `GgufFileInspector.GGUF_MAGIC` — `"GGUF".toByteArray(US_ASCII)` (line 22)
- `GgufModelInspector.GGUF_MAGIC` — `0x46554747` (line 24)
- `GgufHeaderReader.GGUF_MAGIC` — `0x46554747` (line 41)

Same for the 256 MB headroom: `ModelMemoryEstimator.COMPUTE_BUFFER_BYTES` (line 44) and `ModelLoadDiagnostics.MIN_RUN_BYTES` (line 20), each documented as "kept equal to" the other.

The duplication used to have a reason — the deleted comment explained it as unavoidable because "ai-assistant cannot import ai-core's classes." This PR removes that classloader boundary: all five constants now live in `ai-agent-local/…/model/`, same package. What's left is a hand-maintained sync contract ("Keep [GGUF_MAGIC] in step with `GgufModelInspector.GGUF_MAGIC`") with nothing checking it — and the `verifySecureApiKeyStoreParity` Gradle task that used to guard this class of drift was deleted from `ai-core/build.gradle.kts` in this PR.

Suggest one internal `GgufFormat` object in this package holding the magic (both forms if both are genuinely needed) and one `ModelMemory` constant for the 256 MB, so the "keep in sync" comments can go away rather than be trusted.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/backends/BackendRegistry.kt
```diff
@@ -0,0 +1,149 @@
...
+ */
+object BackendRegistry {
+
+    private const val TAG = "$LOG_PREFIX.BackendRegistry"
+
+    /**
+     * Every registered backend, sorted by label so the selector's order is stable across restarts
+     * (the underlying registry is a hash map).
+     *
+     * @return the installed backends; empty when AI Core is absent or no backend registered
```
hal-eisen-adfa: **F15 — this KDoc is written from the perspective of a plugin that no longer exists.**

`BackendRegistry` lives *inside* AI Core, but reads as if it were a consumer talking to AI Core across a service boundary:

- "Everything here comes from **AI Core's** live registry, so no provider is named anywhere in **this plugin**" (line 27)
- "empty when **AI Core is absent** or no backend registered" (line 39) — AI Core can't be absent from itself

After the split this file *is* AI Core's registry view. Rewording it costs nothing now and saves the next reader from concluding there's an IPC hop here that isn't.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/backends/AiBackend.kt
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Deliberately not an enum: backends are contributed by separate plugins now, so AI Core cannot
+ * know the full set at compile time. A new backend plugin stores its own id as the preference
+ * value and needs no change here; the map below only translates the two values that predate that
+ * convention and are already on users' devices.
+ */
+object AiBackend {
+
+    /** SharedPreferences file, owned by the AI Assistant plugin, holding the backend selection. */
+    const val PREFERENCE_FILE = "AgentSettings"
```
hal-eisen-adfa, line 19: **F15 (cont.)** — same stale framing, and one line that's now wrong rather than just dated:

- Line 6: "How AI Core maps the **AI Assistant** backend *setting*" — AI Assistant is deleted by this PR.
- Line 15: `PREFERENCE_FILE` is described as "owned by the **AI Assistant** plugin". It isn't — after `adoptLegacySettings()` this is ai-core's own partition. A reader taking that comment at face value would look for a plugin that isn't in the tree.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/services/LlmInferenceServiceImpl.kt
```diff
@@ -1,6 +1,9 @@
-package com.itsaky.androidide.plugins.aicore
+package com.itsaky.androidide.plugins.aicore.services
 
 import com.itsaky.androidide.plugins.PluginContext
+import com.itsaky.androidide.plugins.PluginLogger
+import com.itsaky.androidide.plugins.aicore.backends.AiBackend
+import com.itsaky.androidide.plugins.aicore.plugin.AiCorePlugin
 import com.itsaky.androidide.plugins.services.LlmInferenceService
 import com.itsaky.androidide.plugins.services.LlmInferenceService.*
 import com.itsaky.androidide.plugins.services.SharedServices
```
hal-eisen-adfa: **F15 (cont.)** — `PluginContext` (line 3) and `SharedServices` (line 9) are dead imports. Grepping this file, neither symbol appears anywhere else — `readSelectedBackendPreference()` now reads `AiCorePlugin.getContext()` instead. Both are compiler warnings; safe to delete.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/backend/LocalLlmBackend.kt
```diff
@@ -76,42 +90,80 @@ class LocalLlmBackend(private val context: PluginContext) : LlmBackend, Cancella
...
+     * engine state or memory headroom.
+     */
+    override fun getSettingsFragmentClassName(): String =
+        "com.itsaky.androidide.plugins.aiagentlocal.settings.LocalLlmSettingsFragment"
+
+    override fun isAvailable(): Boolean {
+        // Check if model is actually configured
+        val prefs = LocalLlmPreferences.of(context)
+
+        val configuredPath = prefs?.getString(LocalLlmPreferences.KEY_MODEL_PATH, null)
```
hal-eisen-adfa: **F15 (cont.)** — `LocalLlmPreferences.of()` returns a non-null `SharedPreferences` (`LocalLlmPreferences.kt:68`), so `prefs?.getString(...)` is a redundant safe call — same at lines 401 and 506. Compiler warns on all three; `prefs.getString(...)` is what's meant.

hal-eisen-adfa (approved):

jatezzz:
Workflow from this PR branch and build against COGO `stage` branch

https://github.com/appdevforall/plugin-examples/actions/runs/32145335224

image

PR #75, opened 2026-08-21
ADFA-5187 | Dynamically size n_ctx based on model metadata and available RAM

jatezzz:
## Description

This PR updates the `llama.cpp` context size (`n_ctx`) to be dynamically calculated at model load, replacing the hardcoded `4096` value. `GgufHeaderReader` parses the `{arch}.context_length` from the GGUF metadata header (`GgufModelInspector`'s own parser is deleted in this PR; it now only classifies chat-vs-embedding from a header it is handed), the header is read once per load and reused by both the embedding guard and the sizing, and the device's available system RAM is checked after the previous model is unloaded. The optimal `n_ctx` is now computed as the minimum of the model's supported context length, the affordable context within the RAM budget, and a sane system ceiling. This change prevents unexpected Android low-memory kills on lower-end devices while unlocking the full context potential for models and devices that can handle > 4096 tokens. The implementation fails open, gracefully falling back to the default `4096` if metadata is missing or unreadable.

## Details

Logic-related changes. Please review the Android logcat during model initialization; you will see logs indicating the parsed model context length, the available RAM snapshot, and the resulting computed `n_ctx` being passed to `LLamaAndroid.load(pathToModel, nCtx)` (`configureContext(...)` is removed in this PR, so the size arrives as a `load` argument rather than a separate configure call).

image

```shell
2026-08-21 11:12:52.065 28532-28789 AiAgentLoc...ma-android com.itsaky.androidide                I  model: loading from /data/user/0/com.itsaky.androidide/files/llm-models/1796906021_675710816_qwen2.5-0.5b-instruct-q8_0.gguf
2026-08-21 11:12:52.346 28532-28789 AiAgentLoc...ma-android com.itsaky.androidide                I  context: using 6 threads (batch=6)
2026-08-21 11:12:52.363 28532-28789 AiAgentLoc...ma-android com.itsaky.androidide                I  context: created with n_ctx = 16384 (requested 16384, model trained for 32768), n_batch = 2048
2026-08-21 11:12:56.272 28532-28789 AiAgentLoc...ma-android com.itsaky.androidide                I  prefill: n_len = 1024, n_ctx = 16384, n_kv_req = 2007
2026-08-21 11:12:56.278 28532-28789 AiAgentLoc...ma-android com.itsaky.androidide                I  prefill: 983 tokens (0 reused from cache) in 1 slice(s) of at most 2048
```

## Also in this PR

- **Sliced prefill** (`llama-android.cpp`). `n_ctx` can now sit far above the batch's fixed capacity, and overrunning that batch corrupts the heap, so the prompt is decoded in slices of `min(the batch's tracked capacity, llama_n_batch(context))`, only the last slice asking for logits. The `prefill: ... in N slice(s) of at most M` line above is the only direct evidence this path ran rather than the old single-batch prefill.
- **`ai-agent-local/libs/v8/llama-v8-release.aar` is regenerated** from the modified `llama-android.cpp` and `LLamaAndroid.kt`. Test a `.cgp` built after this change: an older aar has neither the sized context nor the sliced prefill, and no longer matches the Kotlin API.

## Ticket

[ADFA-5187](https://appdevforall.atlassian.net/browse/ADFA-5187)

## Observation

The fallback mechanism is completely safe and mirrors the previous behavior (defaults to 4096) on any parse failure. The RAM snapshot excludes the currently loaded context to ensure the budget accurately reflects available memory.

[ADFA-5187]: https://appdevforall.atlassian.net/browse/ADFA-5187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ContextSizePolicy.kt
```diff
@@ -0,0 +1,61 @@
...
+        val modelTokens = header.contextLength?.takeIf { it > 0L } ?: return DEFAULT_CONTEXT_TOKENS
+        // Nothing to weigh below the floor, and no reason to price a cache we would not shrink.
+        if (modelTokens <= DEFAULT_CONTEXT_TOKENS) return DEFAULT_CONTEXT_TOKENS
+
+        val perToken = ModelMemoryEstimator.kvBytesPerToken(header)?.takeIf { it > 0L }
+            ?: return DEFAULT_CONTEXT_TOKENS
+
+        // Compute buffers come off the top; goes negative on a short device, which the floor absorbs.
+        val budgetBytes = (availableBytes - ModelMemory.RUN_BUFFER_BYTES) / KV_BUDGET_DIVISOR
+        val affordableTokens = budgetBytes / perToken
```
hal-eisen-adfa: **KV budget is computed from free RAM with no term for the model weights.**

`budgetBytes` reserves `RUN_BUFFER_BYTES` for compute buffers and then halves what's left, but the weights themselves are never subtracted. So the more of free RAM the model occupies, the more aggressively this over-allocates.

Concrete: device reports 3.5 GB `availMem`, user loads a 4.4 GB Q4 7B (28 layers, 4 KV heads x 128 -> ~57 KB/token).

- `budgetBytes = (3.5 GB - 256 MB) / 2 = 1.66 GB`
- `affordableTokens = ~30k` -> capped to `MAX_CONTEXT_TOKENS`
- `n_ctx = 16384` -> **~940 MB of anonymous KV cache**, up from ~235 MB under the old fixed 4096

Those weights are mmap'd and have to page in from the same pool that is now ~700 MB smaller, which worsens exactly the "thrashes for ten minutes then errors" case `MemoryEstimate.loadBytes` documents.

The KDoc on `KV_BUDGET_DIVISOR` justifies the halving as headroom for "the IDE and the app being edited" - that is not headroom for the weights. Consider subtracting the model's on-disk size before dividing (the settings path already has it as `fileInfo.sizeBytes`, and the backend path can stat the resolved file).

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
```diff
@@ -345,24 +385,36 @@ Java_android_llama_cpp_LLamaAndroid_new_1context(JNIEnv *env, jobject, jlong jmo
...
-    const int configured_ctx = g_n_ctx.load();
-    ctx_params.n_ctx = configured_ctx > 0 ? configured_ctx : 4096;
+    int requested_ctx = jn_ctx > 0 ? jn_ctx : DEFAULT_N_CTX;
+
+    // Backstop on Kotlin's number: a misparsed header must not exceed what the model was trained for.
+    const int trained_ctx = llama_model_n_ctx_train(model);
+    if (trained_ctx > 0 && requested_ctx > trained_ctx) {
+        LOGi("context: requested n_ctx %d exceeds the model's trained %d; clamping", requested_ctx, trained_ctx);
+        requested_ctx = trained_ctx;
+    }
```
hal-eisen-adfa: **The clamp is invisible to Kotlin, so the logs and the memory estimate will disagree with what was actually allocated.**

Before this PR `ctx_params.n_ctx` was an unconditional 4096 with no clamp, so this branch is new behavior. For any model trained below 4096 (TinyLlama, phi-1.5, and many small chat models at 2048):

1. `ContextSizePolicy.choose()` returns the 4096 floor - it early-returns `DEFAULT_CONTEXT_TOKENS` whenever `modelTokens <= DEFAULT_CONTEXT_TOKENS`
2. this clamps `requested_ctx` down to 2048
3. `LocalLlmBackend.kt:352` logs `Context size ...: 4096 tokens`, and `ModelMemoryEstimator` prices a 4096-token KV cache
4. but `completion_init`'s `n_kv_req > n_ctx` check runs against 2048

Net effect: with `nlen` up to 1024, a ~1100-token prompt that worked before this PR now throws "Prompt is too long for the model's context size." while every log line says the context is 4096.

`LLamaAndroid.getContextSize()` / `model_n_ctx` already exists and would report the truth, but it has no caller anywhere in the repo. Reading it back after `llama_init_from_model` and logging/storing *that* value would keep the two sides in agreement.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsViewModel.kt
```diff
@@ -276,12 +276,17 @@ class LocalLlmSettingsViewModel(
...
         val estimate = ModelMemoryEstimator.estimate(
             fileSizeBytes = fileInfo.sizeBytes,
-            header = GgufHeaderReader.read { modelFiles.openStream(context, uriString) },
+            header = resolved.header,
+            contextTokens = resolved.contextTokens,
         )
-        // Read last and never cached: the user may have just closed apps to make room.
-        val availableBytes = deviceMemory.availableBytes()
 
         return when (val verdict = ModelMemoryGate.evaluate(estimate, availableBytes)) {
```
hal-eisen-adfa: **The memory gate is now circular: `contextTokens` is derived from `availableBytes`, and the estimate built from it is then compared back against that same `availableBytes`.**

`ModelMemoryGate.Safe` requires `availableBytes >= fileSize + kv + 256 MB`, but `kv` is now up to `MAX_CONTEXT_TOKENS * perToken` (~2 GB for a 7B at ~128 KB/token) instead of a fixed 4096-token cache (~500 MB). Two consequences:

- A 4 GB model on a 5 GB-free device that used to read `Safe` now reads `Risky` and pops the "may not fit" dialog. The larger context the policy just granted is what pushes it over the line.
- The "needs X to run" figure shown to the user is a function of whatever `availMem` happened to be at that instant, so re-selecting the same model after closing an app shows a different number.

Also, the comment on line 281 - "Resolved the same way the load will resolve it, so the warning describes the real allocation" - does not hold. `LocalLlmBackend.kt:312` takes its own `availMem` reading later, after an unload, so the load resolves a *different* context than the one this warning described.

hal-eisen-adfa (commented):
Three findings from a review of the dynamic `n_ctx` change. All three are about the memory accounting around the new context sizing rather than the sizing itself.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ContextSizePolicy.kt
```diff
@@ -0,0 +1,64 @@
...
+        val weightBytes = modelSizeBytes?.takeIf { it > 0L } ?: return DEFAULT_CONTEXT_TOKENS
+        val modelTokens = header.contextLength?.takeIf { it > 0L } ?: return DEFAULT_CONTEXT_TOKENS
+        // Nothing to weigh below the floor, and no reason to price a cache we would not shrink.
+        if (modelTokens <= DEFAULT_CONTEXT_TOKENS) return DEFAULT_CONTEXT_TOKENS
+
+        val perToken = ModelMemoryEstimator.kvBytesPerToken(header)?.takeIf { it > 0L }
+            ?: return DEFAULT_CONTEXT_TOKENS
+
+        // Weights and compute buffers come off the top; negative on a short device, floor absorbs it.
+        val spareBytes = availableBytes - weightBytes - ModelMemory.RUN_BUFFER_BYTES
```
itsaky-adfa: **Signed underflow can turn "no RAM" into the maximum context.**

`availableBytes - weightBytes - RUN_BUFFER_BYTES` is unchecked `Long` arithmetic. When `weightBytes` is very large relative to `availableBytes`, the second subtraction underflows past `Long.MIN_VALUE` and wraps to a large *positive* number, so `budgetBytes`/`affordableTokens` come out huge and `choose` returns `MAX_CONTEXT_TOKENS` instead of the floor.

Concrete: `choose(header, availableBytes = 0L, modelSizeBytes = Long.MAX_VALUE)` -> `0 - Long.MAX_VALUE = Long.MIN_VALUE + 1`, then `- 268435456` wraps to `9223372036586340353`, and the result is `16384` -- the exact opposite of the documented "negative on a short device, floor absorbs it".

`ContextSizePolicyTest.givenAnyInputs_...` already feeds `Long.MAX_VALUE` sizes through this path, but it only asserts the result is *within* the bounds, so it passes on the wrong answer. Reachability today is limited (`modelSizeBytes` comes from `File.length()`), but the function's stated contract is that any input fails safe, and it does not.

A saturating subtraction (or a `coerceAtLeast(0L)` on `spareBytes` before the divide, plus clamping the first subtraction) fixes it.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsViewModel.kt
```diff
@@ -276,12 +280,15 @@ class LocalLlmSettingsViewModel(
         context: Context
     ): Boolean {
         val modelName = fileInfo.displayName
+        // Never cached: the user may have just closed apps to make room.
+        val availableBytes = deviceMemory.availableBytes()
```
itsaky-adfa: **The deliberate "read free RAM last" ordering was inverted.**

Before this change the snapshot was taken *after* the header parse, with the comment `// Read last and never cached: the user may have just closed apps to make room.` Now it is taken first, and the surviving comment keeps only the "never cached" half.

`GgufHeaderReader.read { modelFiles.openStream(context, uriString) }` opens a content URI and does blocking I/O over the model file (up to `MAX_METADATA_BYTES` = 64 MB). For a large model on a slow SAF provider that is a real interval, and the free-RAM figure the warning verdict is judged against is now measured before it rather than after -- exactly the staleness the original ordering was written to avoid.

`header` is not used anywhere else in this function, so there is no reason for the reorder; move the `availableBytes` read back below the `estimate` call.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/GgufHeaderReader.kt
```diff
@@ -23,6 +25,7 @@ import java.io.InputStream
 data class GgufHeader(
     val architecture: String?,
     val blockCount: Long?,
+    val contextLength: Long? = null,
```
itsaky-adfa: **New parameter inserted mid-constructor of a public data class -- positional callers shift silently.**

`GgufHeader` is public, and `contextLength: Long?` lands between `blockCount` and `embeddingLength`. Every neighbouring property is also `Long?`, so any positional construction `GgufHeader(arch, blocks, embedding, heads, headsKv, key, value)` still **compiles** and now binds `embedding` to `contextLength`, `heads` to `embeddingLength`, and so on -- a silent misassignment, not a build error. The same applies to `component3()` in a destructuring declaration and to positional `copy()`.

Every in-repo caller (`GgufHeaderReader.readHeader`, `ContextSizePolicyTest.header`) uses named arguments, so nothing breaks today. But the type is part of this plugin's public surface, and the hazard is free to avoid: append `contextLength` at the end of the parameter list instead.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/backend/LocalLlmBackend.kt
```diff
@@ -326,6 +331,59 @@ class LocalLlmBackend(
...
+     * stored anywhere. [ModelContextResolver] fails open, so this has no failure of its own.
+     *
+     * @param resolvedPath filesystem path to the model, already resolved from any content URI
+     * @param availableBytes free RAM as [availableMemoryBytes] reports it, negative if unknown
+     * @return the context size in tokens to load the model with
+     */
+    private suspend fun resolveContextSize(resolvedPath: String, availableBytes: Long): Int {
+        val resolved = withContext(Dispatchers.IO) {
+            val file = File(resolvedPath).takeIf { it.isFile }
+            ModelContextResolver.resolve(
```
itsaky-adfa: **The GGUF metadata block is now parsed twice on every model load.**

`ensureModelLoaded` already called `GgufModelInspector.classify(resolvedPath)` a few lines above (line 295), which opens the same file and walks the same metadata key/value block to find `general.architecture`. `ModelContextResolver.resolve` here re-opens it and re-walks it from byte 0 to pick up `{arch}.context_length` and the shape values.

On a phone, switching models now means two full metadata parses of a multi-GB file (each bounded at `MAX_METADATA_BYTES` = 64 MB, real blocks a few MB) instead of one. `GgufHeader` already carries `architecture`, so a single `GgufHeaderReader.read` up front could feed both the embedding-model guard and the context sizing -- and would also fix the fact that `classify` does its blocking read *outside* any `Dispatchers.IO` boundary while this one is careful to be inside one.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ContextSizePolicy.kt
jatezzz: @hal-eisen-adfa Fixed in 1706b80 — `choose` now takes `modelSizeBytes` and subtracts the weights before the divisor, so the KV budget is what is left *after* them.
The settings path passes `fileInfo.sizeBytes`, the backend stats the resolved file. Tests pin your 4.4 GB-in-3.5 GB case to the floor, and that a larger model gets less context from the same free RAM.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
jatezzz: @hal-eisen-adfa Fixed in 1706b80 — the native clamp is floored at `DEFAULT_N_CTX`, so a 2048-trained model still gets 4096 and no prompt that fit before this PR regresses.
`LocalLlmBackend.reportEffectiveContextSize` now reads `getContextSize()` back after the load and logs it, warning when it differs from the request — its first caller anywhere.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsViewModel.kt
jatezzz: @hal-eisen-adfa Fixed in 1706b80 — the pre-flight now prices the KV cache at `ContextSizePolicy.DEFAULT_CONTEXT_TOKENS`, never at one derived from the `availMem` it is then compared against.
That breaks the circularity and keeps "needs X to run" stable across two selections of the same model. The comment that claimed the warning described the real allocation went with it.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ContextSizePolicy.kt
jatezzz: @itsaky-adfa Fixed in 5d181b3 — both subtractions are clamped with `coerceAtLeast(0L)`, and a negative `availableBytes` now falls into the unreadable branch rather than being subtracted from.
Added `givenAModelLargerThanFreeRam_...`, `givenNegativeFreeMemory_...` and a monotonicity case; the first two fail against the old policy with `expected:<4096> but was:<16384>`.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsViewModel.kt
jatezzz: @itsaky-adfa Fixed in 5d181b3 — `deviceMemory.availableBytes()` moved back below the header parse and the estimate, and the comment says why it reads last, not just that it is uncached.
`LocalLlmBackend` had the same inversion: it now parses the header first, then unloads, then reads free RAM.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/GgufHeaderReader.kt
jatezzz: @itsaky-adfa Fixed in 5d181b3 — `contextLength` now sits at the end of the parameter list, after `valueLength`.
Added a line to the KDoc saying new properties get appended for exactly this reason, since every one of them is `Long?`.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/backend/LocalLlmBackend.kt
jatezzz: @itsaky-adfa Fixed in 5d181b3 — one `GgufHeaderReader.read` up front, inside `Dispatchers.IO`, feeds both the embedding guard and the context sizing; `ModelContextResolver.resolve` takes the header and is now pure, so free RAM is still read after the unload.
`GgufModelInspector` lost its duplicate parser (~90 lines) and classifies a `GgufHeader`; only the 4-byte `isGguf` check still opens the file.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/backend/LocalLlmBackend.kt
```diff
@@ -287,11 +291,19 @@ class LocalLlmBackend(
...
+            GgufHeaderReader.read { modelFile?.inputStream() } to
+                modelFile?.length()?.takeIf { it > 0L }
+        }
+
         // Guard the chat path against encoder-only embedding models. Running causal generation on
         // one aborts natively (SIGABRT) and takes the IDE down. Classify BEFORE unloading any
         // working chat model, so a wrong selection never tears down a good one. See ADFA-4388.
-        val kind = GgufModelInspector.classify(resolvedPath)
-        if (kind.isEmbeddingOnly) {
+        if (GgufModelInspector.classify(header).isEmbeddingOnly) {
```
hal-eisen-adfa: **The encoder-only crash guard now fails open in more cases than it used to.**

Folding the two metadata parses into one is the right call, but it also widened what can defeat this guard, and that wasn't priced in.

Before `5d181b3`, `GgufModelInspector.classify(path)` used `readArchitecture`, which `return`ed the moment it saw `general.architecture` — conventionally the first KV entry. Nothing later in the metadata block could affect classification, and that parser had no caps at all.

It now consumes `GgufHeaderReader.read`, which must walk **every** KV entry to completion. Any of these makes it return null: an unknown value type, `entryCount > MAX_METADATA_ENTRIES` (4096), an array declaring more than `MAX_ARRAY_ELEMENTS` (2^22), a block over `MAX_METADATA_BYTES` (64 MB), or a key over `MAX_STRING_BYTES` (1 MB). A null header classifies as `ModelKind.UNKNOWN`, and `isEmbeddingOnly` treats that as chat-capable.

So an embedding GGUF (BERT / nomic-bert family) produced by a converter that emits any construct this stricter parser rejects now passes the guard, `llama_decode` runs causal attention on an encoder-only model, and llama.cpp `abort()`s — SIGABRT, no Kotlin `try/catch` can intercept it, whole IDE process gone. That is exactly what ADFA-4388 added this guard to prevent.

One parse in the happy path is still worth keeping. Falling back to an architecture-only scan when `read` returns null would restore the old robustness without reintroducing the second parse for well-formed files.

ai-agent-local/llama-impl/src/main/java/android/llama/cpp/LLamaAndroid.kt
```diff
@@ -231,14 +238,24 @@ class LLamaAndroid : ILlamaController {
...
+    suspend fun load(pathToModel: String, nCtx: Int) {
         withContext(runLoop()) {
             when (threadLocalState.get()) {
                 is State.Idle -> {
                     val model = load_model(pathToModel)
                     if (model == 0L) throw IllegalStateException("load_model() failed")
 
-                    val context = new_context(model)
+                    val context = new_context(model, nCtx)
                     if (context == 0L) throw IllegalStateException("new_context() failed")
```
hal-eisen-adfa: **`load()` leaks the model handle when `new_context` fails — and this PR is what makes that reachable.**

If `new_context` returns 0, the throw on line 259 leaves `model` — an already-mmap'd multi-GB `llama_model` — with no owner: there is no `try` / `free_model` here, and `threadLocalState` stays `Idle`. `LocalLlmBackend` turns the exception into a `ModelLoadException`, the user retries, and every retry mmaps and leaks another model for the process lifetime. `new_batch` / `new_sampler` below have the same problem and leak the context too.

The leak itself predates this PR. What changes is reachability: `n_ctx` was an unconditional 4096, so `llama_init_from_model` returning null was near-impossible. It can now ask for up to 4x the KV cache, sized from an `availMem` snapshot taken *before* `llama.load()` spends seconds mmapping and paging a multi-GB file. That moves this branch from theoretical to plausible on a loaded device — and retrying after a "model failed to load" message is precisely what a user does next.

A `try` / `catch` that frees whatever has been allocated so far before rethrowing would keep the failure recoverable instead of cumulative.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ContextSizePolicy.kt
```diff
@@ -0,0 +1,70 @@
...
+        // Nothing to weigh below the floor, and no reason to price a cache we would not shrink.
+        if (modelTokens <= DEFAULT_CONTEXT_TOKENS) return DEFAULT_CONTEXT_TOKENS
+
+        val perToken = ModelMemoryEstimator.kvBytesPerToken(header)?.takeIf { it > 0L }
+            ?: return DEFAULT_CONTEXT_TOKENS
+
+        // Weights first, then the compute buffers, each clamped at zero rather than left to run
+        // negative: an unclamped Long would underflow on an absurd size and wrap to a huge
+        // positive budget, turning "no RAM at all" into the ceiling instead of the floor.
+        val afterWeights = (freeBytes - weightBytes).coerceAtLeast(0L)
```
hal-eisen-adfa, line 63: **This subtraction and the module's own mmap comments now state opposite memory models.**

Charging the weights against free RAM is what I asked for and the reasoning still holds — the weights page in from the same pool the KV cache is allocated from. But two comments in neighbouring files on the same load path still assert the opposite, and they were not updated:

- `ModelLoadDiagnostics.kt:41-43` — "Weights are mmap'd, so the file need not fit in free RAM — only the KV cache and compute buffers must be resident."
- `ModelMemoryEstimator.kt:8` (`@property loadBytes`) — "the weights. mmap'd, so they need not *fit*"

Neither line is in this diff, which is how they survived. The result is that the next person to touch memory accounting gets contradictory guidance depending on which file they open first, and both of those comments are load-bearing explanations rather than throwaway notes.

Worth reconciling here: either say at this line why the residency assumption differs for context sizing (the snapshot is taken before a load that then pages the weights in), or soften the other two. They should not keep claiming the file need not fit while this line requires exactly that.

One behavioural consequence to confirm is intended rather than incidental: a 4.4 GB Q4 7B advertising 32768 context on a device reporting 3.5 GB `availMem` clamps `afterWeights` to 0 and lands on the 4096 floor. The large models where a bigger context helps most are the ones that never get one. If that is the deliberate conservative outcome, fine — flagging it so it is a decision on the record.

hal-eisen-adfa (commented):
Three follow-ups on the latest round. The seven earlier threads all check out in `5d181b3` and the new tests cover the cases they promised, so nothing here reopens those. Two of these are consequences of the dedup fix and the new variable `n_ctx` respectively; the third is a doc inconsistency the weight-subtraction fix left behind.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/backend/LocalLlmBackend.kt
jatezzz: @hal-eisen-adfa Good catch — added `GgufHeaderReader.readArchitecture`, an architecture-only rescan that runs *only* after `read` returns null and stops at the first `general.architecture`, so no later entry and no entry-count ceiling can defeat the guard.
New `GgufModelInspectorTest` asserts `read` returns null **and** the kind is still `EMBEDDING`, for both an unknown trailing value type and an overstated `entryCount`.
It runs inside `Dispatchers.IO`, so the old outside-IO blocking read is not back either; a well-formed file still costs exactly one parse.

ai-agent-local/llama-impl/src/main/java/android/llama/cpp/LLamaAndroid.kt
jatezzz: @hal-eisen-adfa Fixed — `new_context`/`new_batch`/`new_sampler` are now inside a `try`, and the `catch` frees in reverse allocation order with `free_model(model)` last before rethrowing, so a failed load leaves nothing behind and retries stop accumulating.
Agreed on the reachability point: that is what moved this from theoretical, so it is fixed here rather than deferred.
Note it only reaches the shipped plugin once `scripts/rebuild-llama-aar.sh` regenerates the committed `libs/v8` AAR.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ContextSizePolicy.kt
jatezzz: @hal-eisen-adfa Reconciled both: `ModelLoadDiagnostics.diagnose` is reframed as diagnosing a load that *already* failed, and `ModelMemoryEstimator.loadBytes` keeps the page-thrashing note but no longer reads as licence to treat the weights as free — both now point at `ContextSizePolicy` for why sizing charges them.
`choose`'s KDoc states the residency difference at the source: the snapshot predates the load that pages the weights in.
On your last point — yes, deliberate, and now on the record in that KDoc: a model whose file approaches free RAM gets the 4096 floor. Say so if you would rather charge a fraction of the weights and I will change the policy instead.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ModelMemoryEstimator.kt
```diff
@@ -73,17 +75,31 @@ object ModelMemoryEstimator {
...
     /**
-     * KV cache size for a full context: one key and one value entry per kv head, per layer, per
-     * position. Null unless every value it needs is present and within its ceiling.
+     * KV cache size for a full context of [contextTokens]. Null unless every value it needs is
+     * present and within its ceiling, or the context is not positive.
+     */
+    private fun kvCacheBytes(header: GgufHeader, contextTokens: Int): Long? {
+        if (contextTokens <= 0) return null
+        val perToken = kvBytesPerToken(header) ?: return null
+        return perToken * contextTokens
```
itsaky-adfa: SHOULD FIX: this product can now wrap, and two KDocs in this file still promise it cannot.

The invariant at line 44 ("within them it stays below 2^57") and the one at line 90 ("Stays under 2^44 within the ceilings below, so any context the policy returns fits a Long") were both true while the context was the fixed `RUNTIME_CONTEXT_TOKENS = 4096L`. Now `contextTokens` is an unvalidated `Int` parameter. `MAX_LAYERS`/`MAX_HEADS`/`MAX_WIDTH` bound `perToken` at roughly 2^44, but nothing bounds the multiplier, so 2^44 times a large `Int` overflows and `kvCacheBytes` returns a negative. A negative `runBytes` then makes `MemoryEstimate.totalBytes`'s saturating guard (`loadBytes > Long.MAX_VALUE - runBytes`, line 22) compare against an overflowed bound, and the gate waves through a model it should have rejected - the exact failure the ceilings were added to prevent.

Not reachable today: the settings VM passes 4096 and the backend passes a `ContextSizePolicy.choose` result capped at 16384. So this is latent, the same class as the mid-constructor-parameter hazard accepted last round, and it costs one line: clamp at the top of `kvCacheBytes` with `contextTokens.coerceAtMost(ContextSizePolicy.MAX_CONTEXT_TOKENS)`, or state the caller's obligation in the `estimate` KDoc so the next caller knows it owns the bound.

(Line 90 also says "within the ceilings below" - they are declared above it.)

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/backend/LocalLlmBackend.kt
```diff
@@ -287,11 +291,23 @@ class LocalLlmBackend(
             return // Already loaded
         }
 
+        // One parse of the metadata block per load, feeding both the guard below and the context
+        // sizing after the unload: it sits at the front of a multi-GB file, and a model switch
+        // used to walk it twice.
+        val modelFile = File(resolvedPath).takeIf { it.isFile }
```
itsaky-adfa: SHOULD FIX: `isFile` is a blocking `stat`, and it sits two lines outside the `withContext(Dispatchers.IO)` this same commit added directly below for exactly this reason.

`File.isFile` hits the filesystem. On a slow or unavailable volume - SD card, a removed USB drive, a stale SAF mount - it blocks whatever dispatcher `load` was called on, which is what lines 299-301 exist to avoid. Getting the header read off the caller's thread while leaving the stat on it only narrows the window.

`modelFile` is used by `openModel` (line 298) and inside the block, so the fix needs a small shuffle rather than a straight move: either return it from the first `withContext` alongside the header and size, or drop the local and let `openModel` be `{ File(resolvedPath).takeIf { it.isFile }?.inputStream() }` so every stat happens inside an IO block.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsViewModel.kt
```diff
@@ -276,11 +280,15 @@ class LocalLlmSettingsViewModel(
         context: Context
     ): Boolean {
         val modelName = fileInfo.displayName
+        // The floor is the least the load can use, so also the least this model can cost.
+        val header = GgufHeaderReader.read { modelFiles.openStream(context, uriString) }
         val estimate = ModelMemoryEstimator.estimate(
             fileSizeBytes = fileInfo.sizeBytes,
-            header = GgufHeaderReader.read { modelFiles.openStream(context, uriString) },
+            header = header,
+            contextTokens = ContextSizePolicy.DEFAULT_CONTEXT_TOKENS,
```
itsaky-adfa: SHOULD FIX: this argument *is* the fix for the circular-memory-gate finding, and no test pins it.

There is no `ModelMemoryEstimatorTest`, no `ModelMemoryGate` test and no settings-VM test in this PR. If `contextTokens` here is ever re-derived from free RAM - which is what the original report was about, and the natural thing for a future reader to "improve" it to - the circularity comes straight back and the whole suite still passes. The other two rounds each shipped a failing-then-passing test (`ContextSizePolicyTest`, `GgufModelInspectorTest`); this fix is the odd one out.

One test is enough: assert `ModelMemoryEstimator.estimate(...)` is priced at `DEFAULT_CONTEXT_TOKENS` for a fixed header regardless of what `deviceMemory.availableBytes()` returns.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsViewModel.kt
```diff
@@ -276,11 +280,15 @@ class LocalLlmSettingsViewModel(
         context: Context
     ): Boolean {
         val modelName = fileInfo.displayName
+        // The floor is the least the load can use, so also the least this model can cost.
```
itsaky-adfa: NITPICK: this comment explains the `contextTokens = DEFAULT_CONTEXT_TOKENS` argument three lines down, but sits above the header read, where it reads as a claim about `GgufHeaderReader.read`. Move it onto the argument it justifies.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/GgufModelInspector.kt
```diff
@@ -62,105 +47,36 @@ object GgufModelInspector {
...
-        } catch (_: Exception) {
-            null
-        } ?: return Result(ModelKind.UNKNOWN, null)
+    /**
+     * Classifies the header [GgufHeaderReader] read, so a caller that also needs the rest of it —
+     * the context sizing does — pays for one metadata parse rather than two. Never throws.
+     *
+     * @param header the model's metadata, or null when it could not be read
+     */
+    fun classify(header: GgufHeader?): Result = classifyArchitecture(header?.architecture)
```
itsaky-adfa: NITPICK: this one-argument `classify` has no caller left - not in production code, not in the tests. The load path uses the two-argument overload at line 66, and `ModelLoadDiagnostics` only uses `isGguf`.

That matters a little more than ordinary dead code, because the two-argument form exists precisely because this one fails open when the header has no architecture - the ADFA-4388 case its own KDoc cites. Leaving the weaker variant public means the next caller can reach it by tab-completion and reintroduce the guard hole. Make it `private` and inline it into the overload, or delete it.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ModelContextResolver.kt
```diff
@@ -0,0 +1,44 @@
...
+     * [GgufHeaderReader.read] returns for anything it could not open or parse, and
+     * [ContextSizePolicy.choose] answers its default for one.
+     *
+     * @param header the model's parsed metadata, or null when it could not be read
+     * @param availableBytes free RAM in bytes, or null when it could not be read
+     * @param modelSizeBytes the model file's size in bytes, or null when it could not be read; the
+     *   weights are charged against free RAM before the KV cache gets a budget
+     * @return the context to load with, and the header behind it
+     */
+    fun resolve(
```
itsaky-adfa: NITPICK: `resolve` is a pure pass-through - it forwards all three arguments to `ContextSizePolicy.choose` and pairs the result with the header it was handed. `ModelContextSize` adds only `advertisedTokens`, which is `header?.contextLength` and exists for one log line.

`LocalLlmBackend.resolveContextSize` could call `choose` directly and read `header?.contextLength` where it logs, dropping this file, the data class and `ModelContextResolverTest`. The test is worth a second look either way: it exercises `GgufHeaderReader`'s fail-open behaviour under this type's name, so the wrapper makes the covered surface look broader than it is.

Not blocking - raising it while the file is new and folding it in is cheap.

ai-agent-local/llama-impl/src/main/java/android/llama/cpp/LLamaAndroid.kt
```diff
@@ -367,12 +400,6 @@ class LLamaAndroid : ILlamaController {
             native_configureSampling(temperature, topP, topK)
         }
 
-        @JvmStatic
-        fun configureContext(nCtx: Int) {
```
itsaky-adfa, line 371: SHOULD FIX: the PR description points reviewers at this deleted API.

It asks us to watch logcat for the computed `n_ctx` "being passed to `LLamaAndroid.configureContext(...)` before context creation". This PR deletes `configureContext` and its `native_configureContext` counterpart; the size is now a `load(path, nCtx)` parameter. Anyone reviewing or QA-ing from the description hunts for a call site that no longer exists.

Two behavioural changes are also missing from the body, and QA needs both: the chunked prefill in `completion_init` (batch capacity can now sit well below `n_ctx`, so long-prompt behaviour changed - the `1 slice(s) of at most 2048` log line in the description is evidence of it but the text never mentions it), and the regenerated `libs/v8/llama-v8-release.aar`, without which none of the native or `llama-impl` changes ship at all.

itsaky-adfa (commented):
Round 4 review of `6667a3a`, with a re-check of every issue flagged in rounds 1-3.

**All 10 previously flagged issues are fixed.** Verified independently, not taken from the reply comments:

*Round 1 (hal-eisen-adfa)* - the KV budget now subtracts the model weights before dividing (`ContextSizePolicy.kt:63-64`), with tests pinning both the 4.4 GB-in-3.5 GB floor case and "a bigger model gets less context from the same RAM". The native clamp is floored at `DEFAULT_N_CTX`, so a 2048-trained model still gets 4096 and no previously-working prompt regresses, and `reportEffectiveContextSize` reads `getContextSize()` back and warns on a mismatch. The circular memory gate is broken by pricing the pre-flight KV cache at `ContextSizePolicy.DEFAULT_CONTEXT_TOKENS`.

*Round 2 (itsaky-adfa)* - the signed underflow is closed on both subtractions; re-derived from scratch, both operands are now non-negative so no wrap is representable, and three tests pin it. Free-RAM reading is back below the header parse and the backend reads `availMem` after the unload. `contextLength` moved to the last constructor position with a KDoc rule. The double metadata parse is one `GgufHeaderReader.read` inside `Dispatchers.IO` feeding both the guard and the sizing, and `GgufModelInspector`'s duplicate parser is gone.

*Round 3 (hal-eisen-adfa)* - `readArchitecture` rescans with no entry-count ceiling and returns at the first `general.architecture`; the new `GgufModelInspectorTest` asserts `read` returns null *and* the kind is still `EMBEDDING` for both an unknown trailing value type and an overstated `entryCount`. The partial-load handle leak is fixed on both paths: Kotlin frees sampler -> batch -> context -> model and rethrows, and on the JNI path `throw_java` + `return 0` surfaces at the assignment so `context` stays `0L` and the catch frees the model. Both mmap comments are reconciled.

**Also checked:** the committed `libs/v8/llama-v8-release.aar` is genuinely rebuilt (`native_configureContext` is gone from `libllama-android.so`, `classes.jar` has `load(String,int)` and 4 `free_*` calls with an exception table in `LLamaAndroid$load$3`), `:ai-agent-local:testReleaseUnitTest` passes, every `ThrowNew(FindClass(...))` is converted to `throw_java`, every `completion_init` exit releases `jtext`, no stale callers of the removed APIs remain, and chunked prefill's `chunk_limit = min(2048, llama_n_batch)` keeps `common_batch_add`'s `seq_id[n_tokens]` probe in bounds despite this fork's `new_batch` dropping upstream's `+1` sentinel.

**New findings: none critical or high**, all posted inline. Three SHOULD FIX (a latent overflow two KDocs still deny, an untested fix, a blocking `stat` outside the IO dispatcher), one SHOULD FIX on the PR description, and two NITPICK (a dead fail-open API, a pass-through wrapper) plus a comment-placement nit.

The description also still says the change "extends the `GgufModelInspector` to parse the `<architecture>.context_length`" - that parsing lives in `GgufHeaderReader`, and `GgufModelInspector`'s own parser was deleted in this PR. Worth a pass over the body before merge; QA reads it.

itsaky-adfa (approved):
Approving. All 10 issues from rounds 1-3 are fixed and independently verified (details in my round 4 review above), and nothing in the new findings is critical or high.

@jatezzz please address the four SHOULD FIX items from that review before merge:

1. `ModelMemoryEstimator.kt:84` - clamp `contextTokens`, or move the bound into the `estimate` KDoc as a caller obligation. The overflow is not reachable from today's callers, but the two KDocs currently deny a wrap that is now possible.
2. `LocalLlmBackend.kt:297` - move the `isFile` stat inside an IO block; it is the one piece of blocking filesystem work left on the caller's dispatcher.
3. `LocalLlmSettingsViewModel.kt:288` - add the one test that pins the KV cache to `DEFAULT_CONTEXT_TOKENS` regardless of free RAM, so the circular gate cannot come back unnoticed.
4. PR description - it points reviewers at the deleted `LLamaAndroid.configureContext(...)`, credits `GgufModelInspector` with parsing that now lives in `GgufHeaderReader`, and omits both the chunked prefill and the regenerated `libs/v8/llama-v8-release.aar`. QA works from this text.

The three NITPICK comments are yours to take or leave.

Approving rather than blocking because none of these can produce a wrong result on the shipped paths: fix 1 is latent, 2 is a narrow window on an already-slow operation, 3 is regression insurance, 4 is documentation. I would still rather see 3 and 4 land here than as a follow-up, since both get much harder to justify once the PR is merged.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ModelMemoryEstimator.kt
jatezzz: @itsaky-adfa Fixed in ba1d14b — `kvCacheBytes` clamps with `coerceAtMost(MAX_CONTEXT_TOKENS)` before the multiply, so no caller can wrap the product.
Both KDocs hold again: the ceilings bound `perToken` under 2^44, the clamped product under 2^58, and line 90 now says "declared above".
A new `ModelMemoryEstimatorTest` case fails against the old code with a negative `runBytes`.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/backend/LocalLlmBackend.kt
jatezzz: @itsaky-adfa Fixed in ba1d14b — `openModel` is now `{ File(resolvedPath).takeIf { it.isFile }?.inputStream() }`, and the size comes from `File(resolvedPath).length()` inside the block.
Both stats happen on IO and the local is gone, so nothing crosses the boundary already resolved.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsViewModel.kt
jatezzz: @itsaky-adfa Fixed in ba1d14b — made structural as well as tested: `ModelMemoryEstimator.estimateForSelection(fileSizeBytes, header)` takes no context and no free-RAM argument, so re-deriving it would mean changing a signature.
New `ModelMemoryEstimatorTest` asserts the price is exactly the floor cache, and that it differs from what `choose`'s answer on a roomy device would give.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsViewModel.kt
jatezzz: @itsaky-adfa Fixed in ba1d14b — the rationale moved onto `estimateForSelection`'s KDoc, where the floor is actually chosen; one line stays at the call site.
Nothing sits above the header read pretending to describe it any more.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/GgufModelInspector.kt
jatezzz: @itsaky-adfa Deleted in ba1d14b rather than made private — the overload calls `classifyArchitecture` directly and never needed it.
Its KDoc merged into the survivor, with a line saying there is deliberately no header-only variant, since that is the form that fails open on the ADFA-4388 files.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ModelContextResolver.kt
jatezzz: @itsaky-adfa Folded in ba1d14b — `ModelContextResolver` and `ModelContextSize` are gone; `resolveContextSize` calls `ContextSizePolicy.choose` directly and reads `header?.contextLength` where it logs.
The test is now `ModelContextSizingTest`, exercising read-then-choose under its own name, with an added assert that the header really is null.

ai-agent-local/llama-impl/src/main/java/android/llama/cpp/LLamaAndroid.kt
jatezzz: @itsaky-adfa Agreed on all three counts. The rewritten body replaces `configureContext(...)` with `load(path, nCtx)` and adds bullets for the chunked prefill in `completion_init` and the regenerated `libs/v8/llama-v8-release.aar`.
It also gains a QA section for the >2048-token prompt path and a 2048-trained model landing on 4096.

PR #76, opened 2026-08-21
ADFA-5188 | Enable KV cache quantization and flash attention

jatezzz:
## Description

This PR implements KV cache quantization (`q8_0`) and enables flash attention in the `llama.cpp` context parameters to drastically improve memory efficiency and generation speed.

* **What:** Configures the context parameters `type_k` and `type_v` to use `GGML_TYPE_Q8_0` and enables flash attention (`LLAMA_FLASH_ATTN_TYPE_AUTO`).
* **How:** These settings are exposed through the configure-before-load pattern. If the native context creation fails (e.g., the model's head width is incompatible with the quantized block size), it gracefully falls back to the previous defaults: `f16` KV cache and flash attention disabled.
* **Why:** Storing the KV cache as `q8_0` halves its byte size, allowing for much longer conversations before context is dropped and drastically reducing mid-generation crashes on 4–6 GB RAM devices. Flash attention mitigates generation latency on longer contexts.

## Details

Logs confirming `n_ctx` initialization, cache type used (`q8_0` vs `f16`), and fallback activations.

### Flash Attention Enabled

* Model: [Qwen3 0.6B Q8_0](https://huggingface.co/Qwen/Qwen3-0.6B-GGUF/resolve/23749fefcc72300e3a2ad315e1317431b06b590a/Qwen3-0.6B-Q8_0.gguf)
* Device RAM: ≥8 GB

Screenshot 2026-08-21 at 12 16 48 PM

### Q8_0 KV cache 

* Model: [Llama 3.2 1B Instruct](https://huggingface.co/unsloth/Llama-3.2-1B-Instruct-GGUF/resolve/b69aef112e9f895e6f98d7ae0949f72ff09aa401/Llama-3.2-1B-Instruct-Q4_K_M.gguf)
* Device RAM: ≤3 GB
* Prompt: ~10K Tokens

Screenshot 2026-08-21 at 1 01 13 PM

## Ticket

[ADFA-5188](https://appdevforall.atlassian.net/browse/ADFA-5188)

## Observation

This implementation works in tandem with dynamic `n_ctx` sizing and should be validated alongside ADFA-5187, as the memory measurement relies on both features working concurrently.

https://github.com/appdevforall/plugin-examples/pull/75 Needs to be merged first

[ADFA-5188]: https://appdevforall.atlassian.net/browse/ADFA-5188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
```diff
@@ -389,20 +407,48 @@ Java_android_llama_cpp_LLamaAndroid_new_1context(JNIEnv *env, jobject, jlong jmo
...
+        // can refuse — no block-size constraint on the cache, and no graph for AUTO to fail to
+        // place. It costs the attention speed-up on a model whose only problem was the cache type,
+        // which is the cheaper mistake to make. Kotlin already screens the head width, so getting
+        // here at all means the header and llama.cpp disagreed.
+        LOGe("Context creation failed; retrying at f16 with flash attention off and n_ctx %d",
+             fallback_ctx);
+        ctx_params.type_k = GGML_TYPE_F16;
+        ctx_params.type_v = GGML_TYPE_F16;
+        ctx_params.flash_attn_type = LLAMA_FLASH_ATTN_TYPE_DISABLED;
+        ctx_params.n_ctx = fallback_ctx;
```
hal-eisen-adfa, line 451: **The fallback retries at essentially the same allocation size, so it can't recover from the failure it will most often hit.**

Kotlin sizes `nCtx` against q8_0 and `fallbackNCtx` against f16 out of the *same* RAM budget, so `fallback_ctx * f16_bytes_per_token ≈ requested_ctx * q8_bytes_per_token` — the retry asks the allocator for roughly the same number of bytes that just failed. And when `quantize_kv` is false, `ModelContextResolver` sets `fallbackContextTokens == contextTokens` (both `ContextSizePolicy.choose(header, avail, F16)`) and `type_k`/`type_v` are already F16, so the retry is byte-identical except flash attention is off.

Concretely: device reports ~3 GB free, policy picks `n_ctx = 16384` (~1 GB KV for a 7B GQA model), another app claims memory during the multi-second load, `llama_init_from_model` fails and returns nullptr. The retry allocates ~1 GB again, fails again, and `new_context` throws. On `main` the same device loaded fine because `n_ctx` was always 4096.

Retrying at `DEFAULT_N_CTX` is the one lever that would actually help here, and it's never pulled. Suggest the fallback drop the context to the 4096 floor (or at least halve it) rather than only changing the cache type.

ai-agent-local/llama-impl/src/main/java/android/llama/cpp/LLamaAndroid.kt
```diff
@@ -241,21 +246,31 @@ class LLamaAndroid : ILlamaController {
...
+        fallbackNCtx: Int = nCtx,
+    ) {
         withContext(runLoop()) {
             when (threadLocalState.get()) {
                 is State.Idle -> {
                     val model = load_model(pathToModel)
                     if (model == 0L) throw IllegalStateException("load_model() failed")
 
-                    val context = new_context(model, nCtx)
+                    val context = new_context(model, nCtx, quantizeKv, fallbackNCtx)
```
hal-eisen-adfa: **A failed `new_context` leaks the entire `llama_model`, and this PR makes that path much more likely.**

`load_model()` has already succeeded and returned a valid handle. If `new_context()` throws (via `throw_java`) or returns 0, `load()` propagates without ever calling `free_model(model)`. `threadLocalState` is still `State.Idle`, so `unload()` is a no-op, and no other reference to the handle exists — the weights (hundreds of MB to several GB, plus the mmap) stay pinned for the life of the IDE process.

`LocalLlmBackend` catches the exception and leaves `modelLoaded = false`, so a user who taps retry runs `load_model()` again and leaks a second copy. Since the dominant failure cause here is memory pressure, every leak makes the next attempt fail harder.

Same hole on the `new_batch` (line 276) and `new_sampler` branches — those leak model + context. A `try`/`catch` that frees what's already been allocated before rethrowing would close all three.

Worth noting this interacts with the fallback above: making `n_ctx` large and device-dependent is exactly what turns this from a theoretical path into a reachable one.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ContextSizePolicy.kt
```diff
@@ -34,20 +34,37 @@ object ContextSizePolicy {
...
+        kvType: KvCacheType = KvCacheType.F16,
+    ): Int {
         // Each null is a distinct "we don't know"; all of them mean the same fallback.
         if (header == null || availableBytes == null) return DEFAULT_CONTEXT_TOKENS
         val modelTokens = header.contextLength?.takeIf { it > 0L } ?: return DEFAULT_CONTEXT_TOKENS
         // Nothing to weigh below the floor, and no reason to price a cache we would not shrink.
         if (modelTokens <= DEFAULT_CONTEXT_TOKENS) return DEFAULT_CONTEXT_TOKENS
 
-        val perToken = ModelMemoryEstimator.kvBytesPerToken(header)?.takeIf { it > 0L }
+        val perToken = ModelMemoryEstimator.kvBytesPerToken(header, kvType)?.takeIf { it > 0L }
```
hal-eisen-adfa, line 78: **The KV budget (just below, `(availableBytes - COMPUTE_HEADROOM) / 2`) takes no account of the model weights that the very next step pulls into the same free-RAM pool.**

That `/ 2` is the only allowance for the weights. On the load path in `LocalLlmBackend.ensureModelLoaded` the order is: `unload` → `availableMemoryBytes()` → `refuseBeforeLoad()` (a flat 256 MB floor, nothing model-specific) → `resolveContextSize()` → `llama.load()`. `ModelMemoryGate` never runs on this path at all — it only runs at model-selection time in the settings screen, off a different free-RAM reading. So on a device reporting 3 GB free with a 3 GB q4 model, the policy grants a ~1.37 GB KV budget and picks a 16384-token context, and llama.cpp then allocates that cache *on top of* paging in 3 GB of mmap'd weights. Thrash, or an outright load failure (which lands in the two issues above).

There's also a feedback loop with the settings-time gate: because KV is now grown to ~half of free RAM, `MemoryEstimate.totalBytes` is guaranteed to be at least `fileSize + avail/2`, so any model file larger than half of free RAM is now *always* reported Risky/TIGHT.

Subtracting the model file size from the budget before dividing, and/or running the gate on the load path with the resolved context, would make the number mean what the comment says it means.

ai-agent-local/llama-impl/build.gradle.kts
```diff
@@ -25,6 +25,7 @@ android {
 				arguments += "-DLLAMA_BUILD_COMMON=ON"
 				arguments += "-DGGML_LLAMAFILE=OFF"
 				arguments += "-DCMAKE_BUILD_TYPE=Release"
+				arguments += "-DANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES=ON"
```
hal-eisen-adfa, line 30: **This is a 16 KB page-size ABI migration riding inside a KV-cache PR, and nothing pins it.**

I parsed both committed AARs to confirm: on `origin/main` every `PT_LOAD` segment of `libllama.so` and `libllama-android.so` has `p_align = 0x1000` (4 KB); on this branch they're `0x4000` (16 KB). That's a real native packaging change — it changes how the `.so` must be aligned inside the host APK/`.cgp` zip — and neither commit message ("size n_ctx per model and device", "store the KV cache as q8_0") mentions it, so it won't get reviewed or device-verified as the ABI change it is.

Two asks:
1. Call it out in the PR description / commit message so it gets device-verified deliberately (it's a good change, it just shouldn't arrive silently).
2. `PrebuiltAarAbiTest` asserts ABI presence, required library names, and non-zero size — nothing asserts `p_align`. The next manual run of `scripts/rebuild-llama-aar.sh` without this flag silently reverts to 4 KB and no test fails. An assertion on segment alignment would hold it.

hal-eisen-adfa (commented):
Four things worth a look before this merges — details inline. In short: the native fallback allocates about the same number of bytes as the attempt that just failed, a failed `new_context` leaks the whole model, the KV budget doesn't account for the weights that get mmap'd right after, and there's a 16 KB page-size ABI change riding along unannounced.

itsaky-adfa (commented):
Re-reviewed at 46d69ec against the stacked base `fix/ADFA-5187-dynamic-n-ctx`. The KV quantization arithmetic is sound and I verified it end to end; the findings are concentrated in the native fallback ladder, which is the part nothing exercises.

**Re-check of the previous round** (4 findings from @hal-eisen-adfa, all four threads currently marked resolved):

- *Fallback retries at essentially the same allocation size* (llama-android.cpp) - **addressed, but incompletely**, and the gap is the one IMPORTANT finding in this round. Replied in the thread and reopened it; f570d09 added the 4096-floor retry, but its guard still fires an attempt that needs more memory than the one that just failed in the common case.
- *Failed `new_context` leaks the whole model* - **fixed**, verified at head, not on the reply: `LLamaAndroid.kt:286-293` frees sampler, batch and context in reverse order and then `free_model(model)` before rethrowing, covering the `new_batch` and `new_sampler` branches too.
- *KV budget ignores the weights mmap'd right after* - **fixed**, verified at head: `ContextSizePolicy.kt:82-84` now subtracts `weightBytes` before `RUN_BUFFER_BYTES` and the divisor, and `choose` returns the floor outright when `modelSizeBytes` is null.
- *16 KB page-size ABI change riding along unannounced* - **half fixed**. The requested assertion landed and genuinely pins something: I extracted both AARs and `origin/main` has `p_align = 0x1000` on five of six libraries (libomp.so was already `0x4000`) against `0x4000` on all six here, so the new test fails against the pre-fix artifact rather than passing vacuously. The description half was not done; replied in that thread and reopened it.

**Verification run for this round**

- `../gradlew testDebugUnitTest` at head: 98 tests, 0 failures, including ContextSizePolicyTest 27, KvCacheTypeTest 10, ModelContextResolverTest 7, ModelMemoryEstimatorTest 5, PrebuiltAarAbiTest 5.
- The committed AAR is genuinely rebuilt from the new `.cpp`, not a stale copy - `libllama-android.so` carries the new log strings (`kv cache = %s`, `retrying at f16 with flash attention off`, `retrying at the n_ctx`). Worth stating because a stale AAR would not have failed anything: JNI resolves `new_context` by name alone, so a 2-arg native function would have silently ignored the two new arguments.
- The new ELF offsets in `PrebuiltAarAbiTest` are correct for ELF64 (e_phoff 32, e_phentsize 54, e_phnum 56, p_align 48, magic `0x464C457F` LE), and the probe window is ample - the program-header table here ends at byte 624 of 4096.
- The q8_0 arithmetic in `ContextSizePolicyTest:158` checks out by hand: the same budget gives 5000 -> 4864 tokens at f16 and 9411 -> 9216 at q8_0.
- Overflow reasoning in `kvBytesPerToken` still holds after the refactor: at the declared ceilings, elements peak at 2^43, so the q8_0 intermediate `elements * 34` reaches ~2^48 and the f16 product ~2^58 - both inside a Long, and f16 is indeed the dearer of the two as the doc claims.

**Not a finding, recording it because it was raised and is wrong**: a parallel pass flagged `KvCacheType.supports` deriving head width from `embeddingLength / headCount` as a SIGABRT risk when the guess is wrong. It is not. llama.cpp re-checks against its own `hparams.n_embd_head_k`/`n_embd_head_v` and returns `nullptr` (llama-context.cpp:2311-2327), which is the whole reason AUTO is used over ENABLED. A wrong Kotlin guess costs a failed context creation and a retry, exactly as the comment at :420 claims.

**Device verification is still owed.** Per the repo's CLAUDE.md, `assemblePlugin` succeeding is not verification for these plugins, and I have not run this on a device. Two things need it specifically: the 16 KB page-size change on a 16 KB-page arm64 device, and any rung of the fallback ladder at all.

Severity key: CRITICAL / IMPORTANT / MINOR / NITPICK, one per comment. This repo has no written approve/request-changes rule - CLAUDE.md covers conventions and verification but sets no merge gate, and there is no REVIEW.md or CONTRIBUTING.md - so the default applied: any confirmed IMPORTANT blocks.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
```diff
@@ -389,23 +410,68 @@ Java_android_llama_cpp_LLamaAndroid_new_1context(JNIEnv *env, jobject, jlong jmo
...
+        // f16 with flash attention off is the one configuration nothing here can refuse — no
+        // block-size constraint on the cache, and no graph for AUTO to fail to place. It costs the
+        // attention speed-up on a model whose only problem was the cache type, which is the cheaper
+        // mistake to make. Kotlin already screens the head width, so getting here at all means the
+        // header and llama.cpp disagreed.
+        LOGe("Context creation failed; retrying at f16 with flash attention off and n_ctx %d",
+             fallback_ctx);
+        ctx_params.type_k = GGML_TYPE_F16;
+        ctx_params.type_v = GGML_TYPE_F16;
+        ctx_params.flash_attn_type = LLAMA_FLASH_ATTN_TYPE_DISABLED;
```
itsaky-adfa: MINOR: this fallback turns flash attention off, but it was already on before this PR, so a model whose q8_0 cache llama.cpp refuses ends up slower than it is on main.

`llama_context_default_params()` sets `flash_attn_type = LLAMA_FLASH_ATTN_TYPE_AUTO` (llama-context.cpp:2262 at the pinned c9058a71), and the pre-PR code never touched the field. So main already ran these models with flash attention resolved by AUTO, normally on. After this change such a model runs its whole session with it DISABLED, losing exactly the generation-latency win on long contexts that this PR exists to deliver.

That also makes the PR description's "gracefully falls back to the previous defaults: `f16` KV cache and flash attention disabled" wrong on the second half - disabled was never the previous default.

Only `type_k`/`type_v` need to move here. Leaving `flash_attn_type` at AUTO still cannot abort: with an f16 V cache the `ggml_is_quantized(type_v)` guard at llama-context.cpp:2329 does not apply, and a graph AUTO cannot place throws and is caught into a null return at :2337.

Graded MINOR because reaching this needs Kotlin's head-width screen and llama.cpp's `n_embd_head_k` to disagree, which I could not demonstrate for a specific shipped model.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
```diff
@@ -389,23 +410,68 @@ Java_android_llama_cpp_LLamaAndroid_new_1context(JNIEnv *env, jobject, jlong jmo
...
 
+    LOGi("Creating context: n_ctx = %d (model trained for %d), kv cache = %s", requested_ctx,
+         trained_ctx, quantize_kv ? "q8_0" : "f16");
+
     llama_context *context = llama_init_from_model(model, ctx_params);
+    bool quantized_in_use = quantize_kv;
+
+    // Two unrelated failures land here and want opposite retries: a refused quantized cache is not
+    // a shortage and keeps its long context, while a shortage is answered only by fewer bytes.
+    if (!context && quantize_kv) {
```
itsaky-adfa, line 443: MINOR: nothing exercises either rung of this ladder, so the PR's crash-safety argument rests entirely on code that has never run.

There are no native tests, and neither model in the description can reach it: Qwen3-0.6B (head_dim 128) and Llama-3.2-1B (head_dim 64) both divide 32, so `KvCacheType.supports` returns true and llama.cpp's own block-size check passes. Every screenshot in the description is therefore of the happy path. If the first retry has a bug, it surfaces on a user's device on a model none of us picked.

For what it is worth I did confirm the load-bearing claim in the comment at :420 - the fork of this review could not, so it is worth recording. At c9058a71, `llama_init_from_model` with AUTO plus a quantized `type_k`/`type_v` checks `n_embd_head_k % blck_size` and `n_embd_head_v % blck_size` and returns `nullptr` (llama-context.cpp:2311-2327); the FA-placement failure throws and is caught into `nullptr` at :2337. So AUTO does refuse rather than abort, and choosing it over ENABLED is right.

A single test that loads a small GGUF with a head width not divisible by 32 would pin the rung that matters.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ModelMemoryEstimator.kt
```diff
@@ -58,28 +56,38 @@ object ModelMemoryEstimator {
...
+     * The cache type comes from [ContextSizePolicy.chooseKvCache], which reads only the header, so
+     * naming it here adds no free-RAM term to the estimate.
+     *
      * @param fileSizeBytes the model file's size, or null when it is unknown
      * @param header the model's metadata, or null when it could not be read
      * @return the estimate, or null when there is nothing to base one on
      */
-    fun estimateForSelection(fileSizeBytes: Long?, header: GgufHeader?): MemoryEstimate? =
-        estimate(fileSizeBytes, header, ContextSizePolicy.DEFAULT_CONTEXT_TOKENS)
+    fun estimateForSelection(fileSizeBytes: Long?, header: GgufHeader?): MemoryEstimate? = estimate(
```
itsaky-adfa, line 67: MINOR: the pre-flight memory warning is no longer an upper bound on what the load takes, and it goes quiet in precisely the case the new native fallback exists for.

Pricing the selection estimate at `chooseKvCache(header)` halves the cache term for any quantizable model: for the default 24 x 8 x 128 shape that is 26,112 vs 49,152 bytes per token, about 107 MB vs 200 MB at the 4096 floor. When llama.cpp then refuses the q8_0 cache and `new_context` retries at f16, the load allocates the f16 figure the user was never warned about - so the one path where the warning would have mattered is the one where it was suppressed.

The load path is unaffected (`ModelContextResolver` sizes and loads against the same type), so this is only the settings-screen gate.

Pricing the selection estimate at `KvCacheType.F16` keeps it an upper bound over both outcomes while still costing nothing in accuracy for the common case, since the granted context is the floor either way.

ai-agent-local/llama-impl/src/main/java/android/llama/cpp/LLamaAndroid.kt
```diff
@@ -241,43 +246,50 @@ class LLamaAndroid : ILlamaController {
...
+     *   native side may still refuse it, in which case the load falls back to f16 at [fallbackNCtx]
+     * @param fallbackNCtx context size for that f16 fallback, which the caller sizes against f16's
+     *   own per-token cost; defaults to [nCtx], correct when [quantizeKv] is false
      */
-    suspend fun load(pathToModel: String, nCtx: Int) {
+    suspend fun load(
+        pathToModel: String,
+        nCtx: Int,
+        quantizeKv: Boolean = false,
+        fallbackNCtx: Int = nCtx,
```
itsaky-adfa: MINOR: defaulting `fallbackNCtx` to `nCtx` is the wrong default for the flag it ships beside - it makes the f16 retry ask for more memory than the attempt that just failed.

`nCtx` is sized against q8_0 whenever `quantizeKv` is true. A caller that passes `quantizeKv = true` and omits `fallbackNCtx` therefore hands the native side an f16 fallback at the q8_0 token count, which at roughly 1.9x the bytes per cached token is a strictly larger allocation than the one that returned null. The KDoc's "defaults to [nCtx], correct when [quantizeKv] is false" states the safe half of that and not the unsafe half.

No caller can reach it today: `LocalLlmBackend.kt:344` is the only call site and it passes `fallbackNCtx = contextSize.fallbackContextTokens` explicitly.

Requiring `fallbackNCtx` whenever `quantizeKv` is true - or defaulting it to 0 and letting `clamp_context` map that to `DEFAULT_N_CTX` - removes the footgun without touching the current call.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
```diff
@@ -414,9 +480,11 @@ Java_android_llama_cpp_LLamaAndroid_new_1context(JNIEnv *env, jobject, jlong jmo
         return 0;
     }
 
-    // n_ctx now varies per model and device, so a wrong size is invisible in a report without this.
-    LOGi("context: created with n_ctx = %u (requested %d, model trained for %d), n_batch = %u",
-         llama_n_ctx(context), requested_ctx, trained_ctx, llama_n_batch(context));
+    // n_ctx and the cache type now vary per model and device, so a wrong one is invisible in a
+    // report without this.
+    LOGi("Context created: n_ctx = %u (requested %d, model trained for %d), n_batch = %u, kv cache = %s",
+         llama_n_ctx(context), (int) jn_ctx, trained_ctx, llama_n_batch(context),
```
itsaky-adfa: NITPICK: this line stopped logging the number llama.cpp was actually asked for.

The `requested` field was `requested_ctx` and is now `(int) jn_ctx`, so a non-positive argument prints 0 or a negative while 4096 was configured, and when either fallback fired neither the clamped first request nor the fallback size appears anywhere in the successful-load line - only the raw ask and the final `llama_n_ctx`.

Relatedly, `clamp_context` is now called twice (:414 and :417). A model whose trained context is below both values logs the identical "exceeds the model's trained" line twice, which in a bug report reads as two separate clamps on one load.

Passing `requested_ctx` here, and giving the fallback clamp its own wording, restores the before/after pair the comment above is claiming.

ai-agent-local/llama-impl/src/main/java/android/llama/cpp/LLamaAndroid.kt
```diff
@@ -241,43 +246,50 @@ class LLamaAndroid : ILlamaController {
...
 
                         batch = new_batch(2048, 0, 1)
                         if (batch == 0L) throw IllegalStateException("new_batch() failed")
 
                         sampler = new_sampler()
                         if (sampler == 0L) throw IllegalStateException("new_sampler() failed")
-                    } catch (failure: Throwable) {
-                        // Reverse of the allocation order, and the model last: it owns the rest.
+                    } catch (e: Throwable) {
+                        // Model last: the context borrows from it, so it has to outlive the context.
```
itsaky-adfa: NITPICK: unrelated churn in a hunk this PR only needs for the `new_context` argument list.

`catch (failure: Throwable)` becomes `catch (e: Throwable)`, the comment is rewritten to say the same thing, and the KDoc paragraph that explained the partial-load cleanup is deleted in favour of a shorter inline note. None of it concerns KV quantization, and it costs a reviewer a careful read of the leak-cleanup block - in a file where that exact block was the subject of a finding one round ago - to establish that nothing changed.

Reverting the rename and the two comment edits shrinks this hunk to the `new_context(model, nCtx, quantizeKv, fallbackNCtx)` line and the signature above it.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ModelContextResolver.kt
```diff
@@ -0,0 +1,61 @@
...
+        modelSizeBytes: Long?,
+    ): ModelContextSize {
+        // A quantized cache buys about twice the context, so the type is picked before the size.
+        val kvType = ContextSizePolicy.chooseKvCache(header)
+        return ModelContextSize(
+            contextTokens = ContextSizePolicy.choose(header, availableBytes, modelSizeBytes, kvType),
+            kvType = kvType,
+            // Sized here rather than scaled natively, so the fallback obeys the one policy that
+            // knows the floor, the ceiling and the rounding.
+            fallbackContextTokens = ContextSizePolicy.choose(
```
itsaky-adfa: NITPICK: this second `choose` call is provably redundant on the branch where it matters most.

`chooseKvCache` returns `F16` for every header `Q8_0.supports` rejects, and on that branch these arguments are identical to the call on line 51 - so `fallbackContextTokens == contextTokens` by construction, which is what the `@property` KDoc and `givenAnUnreadableHeader_whenResolving_thenTheFallbackSizeIsTheDefaultToo` both already assert. It is pure and cheap, so nothing misbehaves; the invariant just holds by coincidence rather than by construction.

`fallbackContextTokens = if (kvType == KvCacheType.F16) contextTokens else ContextSizePolicy.choose(...)` makes it structural.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
itsaky-adfa: IMPORTANT: the 4096-floor retry added for this still fires an attempt that needs *more* memory than the one that just failed, in the case your own example lands in.

f570d09 added the second retry, but its guard is `floor_ctx < current_ctx || flash_attn_type != DISABLED`. `ContextSizePolicy.choose` never returns below `DEFAULT_CONTEXT_TOKENS` and native `DEFAULT_N_CTX` is the same 4096, so whenever the policy already landed on its floor - tight RAM, unreadable header, unknown model size, or a model advertising 4096 or less - `floor_ctx == current_ctx == 4096` and only the second disjunct fires. That retry re-requests the identical `n_ctx` with the identical f16 cache, changing exactly one thing: flash attention off.

Flash attention off is not cheaper. In the pinned llama.cpp, `llm_graph_context::build_attn_mha` takes `ggml_flash_attn_ext` only when `cparams.flash_attn` (llama-graph.cpp:1287); the else branch materialises `kq = ggml_mul_mat(ctx0, k, q)` (:1329) and `ggml_soft_max_ext` (:1363), an `[n_kv x n_ubatch x n_head]` F32 tensor in the compute buffer that the FA path never allocates.

So on the 3 GB device in your comment, once the policy has fallen back to the floor: attempt 1 OOMs, and the retry asks the allocator for the same KV cache plus a KQ buffer, fails again, and `new_context` throws - having spent a second multi-second context creation to get there. Your original point survives the fix on that path.

Gating the retry on `floor_ctx < current_ctx` alone, and dropping flash attention only where a quantized cache forced it, fixes both. Reopening.

ai-agent-local/llama-impl/build.gradle.kts
itsaky-adfa: MINOR: ask 2 is done and it pins the right thing; ask 1 is still open.

I checked both committed AARs independently and reached your numbers: on `origin/main` five of six libraries have `p_align = 0x1000` (`libomp.so` was already `0x4000`), and on this branch all six are `0x4000`. So `givenTheCommittedAar_whenInspected_thenEveryLibraryIsAlignedForLargePages` fails against the pre-fix artifact rather than passing vacuously, which is the property that makes it worth having. The ELF offsets it uses are correct for ELF64 and the 4096-byte probe is ample here (the program-header table ends at byte 624).

But the description half was not done. The PR body still describes only KV cache quantization and flash attention, and neither commit subject - "store the KV cache as q8_0 to double the context", "retry a failed context at the 4096 floor" - mentions page size. That was the half aimed at getting the ABI change deliberately device-verified, and a QA reader working from this PR still has no reason to reach for a 16 KB-page device. Per the repo's CLAUDE.md, a green `assemblePlugin` is not verification for that class of change.

Reopening for a line in the description.

itsaky-adfa (changes requested):
Requesting changes on one confirmed IMPORTANT, detailed in the reopened `llama-android.cpp` thread.

The 4096-floor retry that f570d09 added to answer the previous round does not shrink the allocation in the case the previous round was about. Its guard fires on `flash_attn_type != DISABLED` alone, so whenever the policy already returned its floor - and `ContextSizePolicy.choose` never returns below 4096, the same value as native `DEFAULT_N_CTX` - the retry re-requests the identical n_ctx and identical f16 cache with flash attention turned off, which adds the `[n_kv x n_ubatch x n_head]` KQ buffer the FA path never allocates (llama-graph.cpp:1287 vs :1329). On the 3 GB device from that thread it asks for more than the attempt that just OOMd.

To clear this: gate the retry on `floor_ctx < current_ctx` alone, and drop flash attention only where a quantized cache forced it.

Everything else is MINOR or NITPICK and none of it blocks. The KV quantization arithmetic is correct - I re-derived the per-token costs, the overflow ceilings and the two exact-value assertions by hand, and the 98 unit tests pass at head. The three other findings from the previous round are genuinely fixed at head, not just claimed fixed.

Two things still owed before QA, both noted inline: device verification of the 16 KB page-size change on a 16 KB-page device, and any exercise at all of the fallback ladder - no test and neither model in the description can reach it.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
jatezzz: @itsaky-adfa Confirmed and fixed: the guard is now `floor_ctx < current_ctx` alone, so at the floor there is no second attempt at all.
Both rungs also stopped disabling flash attention, so no retry trades a KQ compute buffer for a cache it was not short of.
Native rungs are still device-unverified.

ai-agent-local/llama-impl/build.gradle.kts
jatezzz: @itsaky-adfa Ask 1 is still open — the description body has not been updated yet, and this is the line for it.
`-DANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES=ON` moves every `PT_LOAD` segment from `p_align = 0x1000` to `0x4000`, an ABI packaging change needing verification on a 16 KB-page arm64 device.
Thanks for independently reproducing the five-of-six numbers.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
jatezzz: @itsaky-adfa Fixed: both rungs now move only `type_k`/`type_v`, and `flash_attn_type` stays AUTO throughout.
Verified your reading at c9058a713 — `llama-context.cpp:294-330` resolves AUTO itself and sets `cparams.flash_attn = false`, throwing only when `type_v` is quantized.
The description's "flash attention disabled" wording still needs the same correction.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
jatezzz: @itsaky-adfa Not fixed, and I agree it is the weak spot: there is only a JVM `src/test` target, so pinning a rung needs an instrumented target plus a committed GGUF with a head width indivisible by 32.
I pinned the Kotlin half instead — `givenAQuantizableModel_whenResolving_thenTheFallbackSizeIsTheShorterOne` asserts the f16 fallback is genuinely the shorter size.
Thanks for recording the `llama_init_from_model` confirmation; the comment at :420 now leans on it.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ModelMemoryEstimator.kt
jatezzz: @itsaky-adfa Fixed: `estimateForSelection` now prices at `KvCacheType.F16`, so it is an upper bound over both outcomes including the f16 retry.
`givenAQuantizableModel_whenEstimatingForSelection_thenItIsStillPricedAtF16` replaces the old test and asserts `atQ8 < selection == atF16`.

ai-agent-local/llama-impl/src/main/java/android/llama/cpp/LLamaAndroid.kt
jatezzz: @itsaky-adfa Fixed: the default is now `0`, which `clamp_context` maps to `DEFAULT_N_CTX`, and the KDoc states the unsafe half explicitly.
`LocalLlmBackend.kt:342` still passes `fallbackNCtx` and is unchanged.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
jatezzz: @itsaky-adfa Fixed: the success line logs `requested_ctx` again.
`clamp_context` took a `which` label ("requested" / "fallback"), and the fallback clamp moved inside the retry branch — so a load that never retries logs no clamp at all, and two clamps read as two.

ai-agent-local/llama-impl/src/main/java/android/llama/cpp/LLamaAndroid.kt
jatezzz: @itsaky-adfa Reverted: the `failure` name, the original cleanup comment and the partial-load KDoc paragraph are all back, and the added inline comment is gone.
The hunk is now the signature plus the `new_context(model, nCtx, quantizeKv, fallbackNCtx)` line.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ModelContextResolver.kt
jatezzz: @itsaky-adfa Fixed: hoisted `contextTokens` and made it structural — `fallbackContextTokens = if (kvType == KvCacheType.F16) contextTokens else ContextSizePolicy.choose(..., F16)`.
The invariant now holds by construction rather than by coincidence.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/settings/LocalLlmSettingsViewModel.kt
```diff
@@ -280,7 +280,8 @@ class LocalLlmSettingsViewModel(
     ): Boolean {
         val modelName = fileInfo.displayName
         val header = GgufHeaderReader.read { modelFiles.openStream(context, uriString) }
-        // Prices at the floor by construction — it takes no context and no free-RAM figure at all.
+        // Prices at the floor by construction — it takes no context and no free-RAM figure at all,
+        // and names the cache type the load will pick, which depends only on the header.
```
itsaky-adfa, line 284: MINOR: this comment now claims the opposite of what `estimateForSelection` does.

At head it hardcodes `KvCacheType.F16` (`ModelMemoryEstimator.kt:71`) and no longer calls `chooseKvCache`, so it does not name the type the load will pick - for any quantizable model the load picks q8_0 while this prices f16. `7d51871` changed the function and left behind the comment `c45cce1` added, in the one spot a reader looks to find out which type the pre-flight charges.

Drop the second clause, or say it prices at f16 deliberately as the dearer of the two.

```suggestion
        // and prices it at f16, the dearer of the two types, whatever the load will pick.
```

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
```diff
@@ -389,23 +411,62 @@ Java_android_llama_cpp_LLamaAndroid_new_1context(JNIEnv *env, jobject, jlong jmo
...
+    const int requested_ctx = clamp_context(jn_ctx, trained_ctx, "requested");
+    const bool quantize_kv = jquantize_kv == JNI_TRUE;
+
+    // AUTO rather than ENABLED: it is AUTO that makes llama.cpp validate a quantized cache against
+    // the model's head width and refuse it by returning null. ENABLED skips that check and aborts
+    // inside ggml instead, taking the IDE down with it.
+    ctx_params.flash_attn_type = LLAMA_FLASH_ATTN_TYPE_AUTO;
+    if (quantize_kv) {
+        // A quantized V cache is only defined with flash attention, which AUTO may still refuse.
+        ctx_params.type_k = GGML_TYPE_Q8_0;
```
itsaky-adfa, line 424: MINOR: nothing pins the committed AAR to this file, and CI never compiles it.

`ai-agent-local/settings.gradle.kts:41` includes `:llama-impl` only when the llama.cpp submodule is checked out, and `build-plugins.yml`'s `actions/checkout@v4` requests no submodules - so a normal build consumes `libs/v8/llama-v8-release.aar` and never builds this `.cpp`. An edit here without a `scripts/rebuild-llama-aar.sh` run therefore ships as a silent no-op that no test and no workflow can catch. That is a sharp edge on a PR whose review rounds have all landed in this file.

I confirmed by hand that `7d51871` did rebuild it, so this is not a live defect. `PrebuiltAarAbiTest` already asserts properties of that same binary; one marker-string assertion there would automate the check.

ai-agent-local/src/main/kotlin/com/itsaky/androidide/plugins/aiagentlocal/model/ModelMemoryEstimator.kt
```diff
@@ -58,28 +56,39 @@ object ModelMemoryEstimator {
      * the policy granted the very thing that trips the warning, and moves the "needs X to run"
      * figure between two selections of the same model (ADFA-5187).
      *
+     * Priced at [KvCacheType.F16] and not at the type the load will pick: f16 is the dearer of the
+     * two and the type the native fallback drops to, so this stays an upper bound over both
+     * outcomes — including the refused-quantized-cache path the warning exists for (ADFA-5188).
```
itsaky-adfa, line 61: NITPICK: "an upper bound over both outcomes" reads as an upper bound on the load, which this number is not.

It bounds the cache *type* at a fixed context. The paragraph directly above calls the same figure "the least any load can use, so the least this model can cost", and the context term is genuinely unbounded here: a roomy device granted 14080 q8_0 tokens allocates about 368 MB of cache against the 201 MB this reports. That is deliberate - it is ADFA-5187's anti-circularity rule - which is exactly why the looser phrasing is worth tightening.

`ModelMemoryEstimatorTest`'s own KDoc already says it precisely: "an upper bound over both cache types".

itsaky-adfa (commented):
Re-reviewed at `7d51871` against the stacked base `fix/ADFA-5187-dynamic-n-ctx`. The IMPORTANT from the last round is fixed, so nothing here blocks: two MINORs, one NITPICK, and two reopened threads that are both asks about the PR description rather than the code.

`plugin-examples` has no written approve/request-changes rule, so the default applied and MINOR does not block. One thing to flag: my earlier `CHANGES_REQUESTED` still stands on this PR and a `COMMENT` does not lift it. Say the word and I will approve, since the finding it was raised for is resolved.

## Previous rounds, verified at head rather than from the replies

Round 1 (@hal-eisen-adfa):

- Fallback retried at essentially the same allocation size - **FIXED**. The floor rung is gated on `floor_ctx < current_ctx` alone now, so it fires whenever the current context is above 4096. Worked the roomy case through with the test header: 14080 q8_0 (368 MB) -> 7424 f16 (365 MB) -> 4096 f16 (201 MB). The ladder finally has a rung that shrinks.
- Failed `new_context` leaked the whole `llama_model` - **FIXED**. The `catch` frees sampler, batch and context, then `free_model(model)`, then rethrows.
- KV budget took no account of the model weights - **FIXED**. `ContextSizePolicy.choose` subtracts `weightBytes` and `RUN_BUFFER_BYTES` before halving.
- 16 KB page-size ABI migration unpinned - **PARTLY FIXED**, thread reopened. The test is done and genuinely pins; the body line is not.

Round 2 (mine):

- Floor retry still asked for more memory than the attempt that just failed (IMPORTANT) - **FIXED**, as above.
- Fallback turned flash attention off - **CODE FIXED**, thread reopened for the description half only.
- Neither native rung is exercised by a test - **NOT FIXED**. Deferral accepted with a reason; details in the thread. Worth a follow-up ticket, not a blocker here.
- Pre-flight estimate was no longer an upper bound - **FIXED**. `estimateForSelection` hardcodes `KvCacheType.F16`, and `givenASelectedModel_whenEstimating_thenTheCacheIsPricedAtTheFloorContext` asserts against `F16_PER_TOKEN`, so a revert to `chooseKvCache` fails the suite.
- `fallbackNCtx` defaulted to `nCtx` - **FIXED**. The default is `0`, which `clamp_context` maps to `DEFAULT_N_CTX`.
- Success line logged `jn_ctx` instead of the clamped ask - **FIXED**, logs `requested_ctx`.
- Unrelated churn in the leak-cleanup hunk - **FIXED**, reverted; the hunk is now the signature plus the `new_context` call.
- Redundant second `choose` call - **FIXED**, structural on `kvType == F16` now.

## What I checked independently

- **The committed AAR really was rebuilt from the C++ at head.** Every added format string (`flash attention = %s`, `retrying at f16 with n_ctx %d`, `context: %s n_ctx %d exceeds`) is in `libllama-android.so` at `7d51871` and absent at `46d69ec`; every removed one is gone. Only that library changed - the other five are byte-identical, consistent with the llama.cpp submodule pointer not moving.
- **The fallback design holds at the llama.cpp level.** `libllama.so` carries both `quantized V cache was requested, but this requires Flash Attention` and `%s: failed to initialize the context: %s`, so a refusal reaches the retry as a null return rather than propagating out of `llama_init_from_model`.
- **Nothing on the Kotlin side budgets prompts against the requested context.** `reportEffectiveContextSize` reads the native `n_ctx` back and warns when it differs, and no other caller stores `contextTokens` - so a shortened f16 fallback cannot desync prompt limits.
- **The new tests are not vacuous.** For the test header q8_0 costs 26112 B/token against f16's 49152, and the resolver returns 14080 vs 7424, so `fallbackContextTokens < contextTokens` pins a real gap. `assertNotEquals` and `atQ8` are both still live assertions.

## Coverage of this review

No device run, so per this repo's own rule I am not calling the change verified: the native ladder has still never executed here, and both models in the description take the happy path (head_dim 128 and 64 both divide 32). The `/code-review` pass I kicked off resolved "76" against the wrong repository - it returned findings about `deploy-to-firestore-app-distribution.yml` and `gradle.properties`, which are not in this PR - so its output was discarded in full and every finding above is hand-derived from the diff at `7d51871`.

This cannot merge before #75; the base is still that branch, and it is currently level with `main`.

ai-agent-local/llama-impl/build.gradle.kts
itsaky-adfa: @jatezzz Ask 1 is still open at 7d51871 - the body carries no line about the page-size change. I searched it for `FLEXIBLE_PAGE_SIZES`, `16 KB`, `page size` and `p_align`: no match.

Ask 2 I re-confirmed independently rather than from the reply. `origin/main`'s AAR has `p_align = 0x1000` on five of six libraries (`libomp.so` was already `0x4000`); this branch has `0x4000` on all six. So `givenTheCommittedAar_whenInspected_thenEveryLibraryIsAlignedForLargePages` genuinely fails against the pre-fix artifact rather than passing either way.

Reopening for the body line only. The text you drafted reads well as-is.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
itsaky-adfa: @jatezzz The code half is fixed, and I confirmed it in the shipped binary rather than only the source: both `flash_attn_type = LLAMA_FLASH_ATTN_TYPE_DISABLED` assignments are gone, `libllama-android.so` at 7d51871 contains `retrying at f16 with n_ctx %d`, and the old `...with flash attention off and n_ctx %d` string is absent from it.

The description half is still open. The body at 7d51871 still says the fallback drops to "`f16` KV cache and flash attention disabled", which the commit deliberately stopped doing. It also still frames the PR as enabling flash attention, and `llama_context_default_params()` already set AUTO before this PR - `type_k`/`type_v` are the only real change.

Reopening for the body.

ai-agent-local/llama-impl/src/main/cpp/llama-android.cpp
itsaky-adfa: @jatezzz Accepting the deferral - an instrumented target plus a committed GGUF with a head width indivisible by 32 is more than this PR should carry, and I checked that the Kotlin substitute is not vacuous: for that header the resolver returns 14080 q8_0 against 7424 f16, so `givenAQuantizableModel_whenResolving_thenTheFallbackSizeIsTheShorterOne` pins a real gap.

One thing worth recording: the ladder is at least reachable in the shipped binary. `libllama.so` carries both `quantized V cache was requested, but this requires Flash Attention` and `%s: failed to initialize the context: %s`, so llama.cpp converts that throw into the null return the retry keys on instead of propagating it out of `llama_init_from_model`. That does not exercise either rung, but it rules out the way the design could have been dead on arrival.

Leaving this resolved - better as a follow-up ticket than a blocker on this PR.

itsaky-adfa (approved):
Approving - this lifts my earlier `CHANGES_REQUESTED`. The floor retry no longer fires an attempt larger than the one that just failed, and I verified the fix at `7d51871` rather than from the reply: the guard is `floor_ctx < current_ctx` alone, and the roomy case now steps 14080 q8_0 (368 MB) -> 7424 f16 (365 MB) -> 4096 f16 (201 MB), so the ladder finally has a rung that shrinks.

Nothing outstanding blocks the merge. Before it goes in, two things are just description edits and cost nothing:

- Add the 16 KB page-size line to the body (the reopened `build.gradle.kts` thread) - the text you drafted is fine as-is.
- Correct the body's "falls back to `f16` KV cache and flash attention disabled", which this PR deliberately stopped doing, and the framing of it as enabling flash attention when `llama_context_default_params()` already set AUTO.

The two MINORs and the NITPICK on this round are safe to merge as they stand; the stale `LocalLlmSettingsViewModel` comment has a one-click suggestion. The untested native rungs stay a known residual - better as a follow-up ticket than a blocker here.

Note this still cannot merge before #75.

PR #78, opened 2026-08-24
ADFA-5189 | Package for arm64-v8a only

jatezzz:
## Description

This PR limits the native library packaging for the `ai-agent-local` plugin to `arm64-v8a` only. Since the IDE extracts a single ABI and CoGo does not ship an x86 build, packaging other ABIs adds unnecessary download weight. It also implements build-time validation in Gradle and a unit test to prevent regressions if the AAR is improperly regenerated or filters are dropped.

## Details

* Added `arm64-v8a` to the `abiFilters` in the Gradle configurations.

* Added an `assemblePlugin` validation step in `build.gradle.kts` to ensure the resulting `.cgp` package strictly contains the `arm64-v8a` ABI, failing the build otherwise.

* Added `PrebuiltAarAbiTest.kt` to verify that the committed AAR carries the correct native libraries.

* Updated documentation to reflect the `arm64-v8a` 64-bit ARM requirement and the reduced plugin size.

### Before

before

### After

after

https://github.com/user-attachments/assets/eda41c56-6975-4fb0-a1a7-a87280b5726a

## Ticket

[ADFA-5189](https://appdevforall.atlassian.net/browse/ADFA-5189)

## Observation

The new validation step runs as the last action of the assemble task rather than a finalizer to ensure any compile errors are not buried under a "no .cgp found" failure in CI.

[ADFA-5189]: https://appdevforall.atlassian.net/browse/ADFA-5189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ai-agent-local/build.gradle.kts
```diff
@@ -74,6 +83,78 @@ dependencies {
...
+                val dir = pluginDir.get().asFile
+                val artifacts = dir.listFiles()
+                    ?.filter { it.isPluginArtifactFor(isDebug) }
+                    ?.sortedBy { it.name }
+                    .orEmpty()
+                if (artifacts.isEmpty()) {
+                    throw GradleException("$assembleTaskName produced no .cgp under $dir.")
+                }
+
+                artifacts.forEach { cgp ->
```
Daniel-ADFA, line 132: Is this necessary, given ` ndk {
            abiFilters += listOf("arm64-v8a")
        }` above?

jatezzz: @Daniel-ADFA  Yes, it acts as a defensive guardrail. The prebuilt `llama-v8-release.aar` contains four ABIs (`arm64`, `v7a`, `x86`, `x86_64`). While `abiFilters` handles the actual stripping during the build, this check ensures that if the filter is accidentally removed in the future or if an `AGP` update breaks the filtering, the build fails explicitly. Otherwise, we would silently ship a bloated `.cgp` with useless `x86`/`v7` libs.

Daniel-ADFA (approved):

PR #82, opened 2026-09-01
ADFA-5308 | Sync run_app wait and anchor build output on first error

jatezzz:
## Description

**What:** Updated the AI agent's build tools to provide accurate, synchronous build feedback. `run_app` now waits for the build to complete, and `read_build_output` captures the exact compiler errors instead of just the log tail.

**How:**
* Modified `RunAppHandler` to suspend using `withTimeoutOrNull` and `suspendCancellableCoroutine` (up to a 10-minute timeout) while waiting for the `BuildAndLaunchCallback`, rather than reporting immediate success.
* Removed the dead tooling-server retry loop.
* Updated `ReadBuildOutputHandler` to scan for specific compiler/Gradle error markers and window 8000 characters starting from the first error line.
* Scoped `ChatViewModel` at the plugin level rather than the fragment level so an ongoing 10-minute build wait isn't canceled if the user switches bottom-sheet tabs.

**Why:** The agent requires precise compiler feedback to iterate and self-correct. Previously, reporting success before the build ran meant the agent never learned if its code was broken. Furthermore, truncating the log to the very end meant the agent only saw the closing Gradle summary rather than the actual unresolved references.

## Details

* Logic-related: Added extensive unit tests (`ReadBuildOutputHandlerTest`, `RunAppHandlerTest`, and `ChatViewModelStoreTest`) verifying 8000-character budget truncation, error-anchored windows, missing callbacks, and view-model retention across tab switches.

## Ticket

[ADFA-5308](https://appdevforall.atlassian.net/browse/ADFA-5308)
Parent [ADFA-5216](https://appdevforall.atlassian.net/browse/ADFA-5216)

## Observation

* The `ChatViewModel` now outlives the `ChatFragment`. This was necessary to ensure that `run_app` wait loops aren't killed simply because the user navigated to another tab. `clearErrorState()` and `persistState()` lifecycle hooks were updated accordingly.

Depends on https://github.com/appdevforall/CodeOnTheGo/pull/1763

[ADFA-5308]: https://appdevforall.atlassian.net/browse/ADFA-5308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/models/AgentState.kt
```diff
@@ -77,6 +77,21 @@ sealed class AgentState {
...
+    get() = when (this) {
+        is AgentState.Idle -> "Idle"
+        is AgentState.Initializing -> "Initializing"
+        is AgentState.Thinking -> "Thinking"
+        is AgentState.Executing -> "Executing(step ${currentStepIndex + 1}/$totalSteps $description)"
+        is AgentState.Processing -> "Processing"
+        is AgentState.Cancelling -> "Cancelling"
+        is AgentState.Error -> "Error"
+    }
+
```
Daniel-ADFA, line 79: please move user facing strings  to strings.xml

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/handlers/ReadBuildOutputHandler.kt
```diff
@@ -35,24 +43,31 @@ class ReadBuildOutputHandler(
...
-                // Truncate to last 2000 chars to avoid overwhelming the LLM
-                val truncated = if (output.length > 2000) {
-                    "...[truncated]...\n" + output.takeLast(2000)
-                } else {
-                    output
-                }
-
-                Log.d(TAG, "Read ${truncated.length} chars of build output")
+                val window = windowFor(output)
+                Log.d(TAG, "Read ${window.text.length} chars, anchored=${window.anchoredOnError}")
```
Daniel-ADFA: prefer context.logger or clean up logs

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/models/AgentState.kt
jatezzz: @Daniel-ADFA Good catch — the real offenders were `formattedProgress`/`formattedTiming`, now rendered from `state_executing` + `time_format_*` (which already existed unused since #65), so durations follow the device locale too.
`traceLabel` stayed in code: it is logcat only and never reaches the UI — the KDoc now says so.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/handlers/ReadBuildOutputHandler.kt
jatezzz: @Daniel-ADFA Did both: every `Log.d/i/w` here duplicated an `AgentTrace` line, so they are gone (with `TAG` and the import) and the rejections are `AgentTrace.refusal`.
Exception paths use `pluginContext.logger.error(msg, e)` — `AgentTrace` gates previews on `BuildConfig.DEBUG` and would drop the stack trace in release.
Same cleanup applied to `RunAppHandler`.

Daniel-ADFA (approved):

PR #85, opened 2026-09-03
ADFA-5254 | Tag AI network sockets with TrafficStats

jatezzz:
## Description

Tag OpenAI and Gemini network operations using Android `TrafficStats` so AI traffic can be attributed to inference and model catalog requests.

The changes add shared traffic-tag helpers that set the appropriate thread stats tag before socket creation and always clear it in a `finally` block. OpenAI POST requests are tagged as inference and GET requests as catalog traffic, while Gemini tags inference requests, SSE streaming, and the full paginated model catalog flow.

This prevents StrictMode untagged socket violations and enables Android network statistics tools to distinguish AI inference usage from routine catalog requests.

## Details

* Added dedicated inference and catalog traffic tags for both AI agent modules.
* Wrapped OpenAI POST and GET request lifecycles with the appropriate traffic tags.
* Tagged Gemini standard inference and SSE streaming requests as inference traffic.
* Tagged the complete Gemini paginated model catalog loop as catalog traffic.
* Added unit tests verifying that tags are applied, cleared after successful execution, cleared after exceptions, and remain distinct and non-zero.
* Supports validation through StrictMode and `dumpsys netstats` network usage reporting.

Screenshot 2026-09-01 at 10 46 01 AM

## Ticket

[ADFA-5254](https://appdevforall.atlassian.net/browse/ADFA-5254)
Parent: [ADFA-5251](https://appdevforall.atlassian.net/browse/ADFA-5251)

## Observation

Traffic tags are thread-local and are explicitly cleared to prevent tags from leaking into reused `Dispatchers.IO` threads. Network accounting can remain approximate when persistent or pooled sockets retain the tag assigned when the socket was originally created.

[ADFA-5254]: https://appdevforall.atlassian.net/browse/ADFA-5254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Daniel-ADFA (approved):
Approving. Every socket-creating site in both plugins (`OpenAiHttpClient.open()` via `post`/`get`, and the three `GeminiBackend` connections) is wrapped, nothing in plugin-api/libs/metadata changes, and `testDebugUnitTest` passes for both modules locally (CI does not compile the plugins).

Nits, none blocking:

- `OpenAiHttpClient.kt:16-18`: the host installs StrictMode only in debug builds (`BuildConfig.DEBUG && !isPardonEnabled`), not "process-wide".
- `NetworkTags.kt:27-33` (both copies): save and restore the previous tag (`getThreadStatsTag()` / `setThreadStatsTag(prev)`) instead of clearing to -1, as the host's `ListenerState` does; harmless today, but a future nested tag would be clobbered.
- `NetworkTags.kt:6-7`: drop "following the host's convention"; the host mixes ASCII (`JDWL`) and hex (`0xC0DE`) tags.
- `NetworkTags.kt` and `NetworkTagsTest.kt` are byte-identical across the two plugins; a `withTrafficTag` helper in plugin-api is the real fix. Worth a follow-up ticket rather than fixing here.
- `givenTheDeclaredTags_whenInspected_thenTheyAreDistinctAndNonZero` only asserts on literals; optional to drop.

PR #87, opened 2026-09-05
Publish addons to Cloudflare R2; stop using GitHub artifact storage

hal-eisen-adfa:
Implements `docs/superpowers/plans/2026-09-04-addon-distribution.md`, which implements the design and PRD beside it.

## What this does

**Ends the artifact quota problem.** The `publish` matrix in `build-plugins.yml` moved ~15 GB per run to produce nothing new, and the bundle upload stored a second copy. Both are gone. No `upload-artifact` or `download-artifact` step remains anywhere.

**Ends the GreenGeeks dependency.** The scp block, SSH key handling, remote MD5 check, filename staging loop, and the whole deploy job are deleted. Addons now go straight from the build to Cloudflare R2.

**Adds a tool that owns addon identity.** `tools/addons/` derives names from directory names, builds source tarballs, generates a schema-validated catalog, and uploads. It runs offline against fixtures, so naming and metadata are enforced on every pull request rather than discovered at release time.

**Normalizes the addons.** 23 renamed to the convention and moved under `plugins/`, with `templates/`, `snippets/`, and `code-actions/` beside it. The `ai-*` addons and `cotg-ndk` are untouched and listed in `tools/addons/skip.txt` with reasons.

## Verified, not asserted

- A source tarball extracts and **builds** to a real `.cgp` with no repository checkout.
- All 23 addons build; all 23 tarballs generate with correct per-addon jar derivation.
- 97 objects published to `staging/local-verify/` on the live bucket. Correct content types, `content-disposition: attachment` on binaries, CORS on the catalog.
- The gallery renders all 23 addons in a browser; a description page renders with injected chrome.
- On an emulator: the renamed addon loads and its tooltips register under the unchanged `plugin.id`.
- `plugin.id`, `namespace`, and `applicationId` compared before and after for all 23: unchanged.
- 42 tests passing; `addons check` exits 0.

## Notable findings

- Refreshing three of the five shared jars automatically was attempted and **reverted as unsafe** — a name search finds build-logic's 21 KB `common.jar` where `libs/common.jar` is 355 KB. The script now reports `NOT REFRESHED` rather than guessing.
- The `pre-push` hook hung every push: git feeds it refs on stdin and `uv` inherited that stdin. Fixed.
- `minAppVersion` is derived from manifests, not hand-written. Ten addons declare a real `YY.ww` minimum; thirteen carry a legacy placeholder and report none rather than an invented value.

## After merge

Run **Publish addons** with `staging: true` — the last plan step, which cannot run before the workflow reaches the default branch.

https://claude.ai/code/session_019Y2YZweBXuLWdBkbqGEUHq


jatezzz (commented):
Automated code review (medium effort), scoped to `.github/workflows/*.yml`, `site/*` and `tools/addons` as requested. 9 findings inline, most severe first.

Verified clean along the way: all 78 tool tests pass; `addons check` is clean across all 23 addons; `addons tarball` builds all 23 and is byte-reproducible across runs; the `../../libs/` depth was updated consistently in every moved plugin; the tarball layout makes `cd plugins/<Addon> && ../../gradlew assemblePlugin` resolve; no tracked `local.properties` or other secret-shaped file reaches an archive; `#active[hidden]` beats `#active{display:flex}` on specificity.

.github/workflows/update-libs.yml
```diff
@@ -90,158 +90,11 @@ jobs:
...
-          if [ $failures -gt 0 ]; then
-            echo "ERROR: $failures file(s) failed MD5 verification"
-            exit 1
-          fi
-
-      - name: Cleanup SSH key
-        if: always()
-        run: |
-          rm -f ~/.ssh/id_rsa ~/.ssh/known_hosts || true
+          files: 'plugins/*/build/plugin/*.cgp'
```
jatezzz, line 100: @hal-eisen-adfa **HIGH — narrowing this glob to `plugins/*` silently drops 7 addons from every release.**

`scripts/update-libs.sh` still builds every module `addons discover --include-skipped` returns, which includes the seven root-level addons: `ai-agent-gemini`, `ai-agent-local`, `ai-agent-mcp`, `ai-agent-openai`, `ai-core`, `ai-literacy-course`, `cotg-ndk`. This glob no longer matches them, `tools/addons/skip.txt` excludes them from the R2 publish, and the GreenGeeks deploy job that used to ship them is deleted in this PR.

Net effect after merge: those `.cgp` files are built on every libs refresh and then published **nowhere** — no release asset, no website, no gallery. `fail_on_unmatched_files: true` won't catch it, because `plugins/*` does match.

This also contradicts the PR's own design doc, `docs/superpowers/specs/2026-09-04-addon-distribution-design.md:55`: "`update-libs.yml` GitHub Release step — **Untouched**, by instruction. Releases keep shipping module-named `.cgp` files as they do now."

Suggest restoring `'*/build/plugin/*.cgp'` alongside the new path, or listing the root-level addons explicitly.

.github/workflows/build-plugins.yml
```diff
@@ -54,94 +53,23 @@ jobs:
...
-            src="$(ls "${module}/build/plugin/"*.cgp 2>/dev/null | grep -v -- '-debug\.cgp$' | head -n1)"
-            if [ -z "$src" ]; then
-              echo "ERROR: no release .cgp found under ${module}/build/plugin/ (did assemblePlugin run?)"
-              exit 1
+          set -euo pipefail
+          all="$(uv run --directory tools/addons addons --root "$GITHUB_WORKSPACE" discover)"
+          # A single-plugin run only builds that module, so the others have no
+          # .cgp by design and must not be reported as missing.
+          if [ -n "$ONLY_PLUGIN" ]; then
+            all="$(printf '%s\n' "$all" | grep -Fx -e "$ONLY_PLUGIN" -e "plugins/$ONLY_PLUGIN")"
```
jatezzz: @hal-eisen-adfa **MEDIUM — a single-addon run of any skipped addon fails *after* a fully successful build.**

`all` is populated from `addons discover` **without** `--include-skipped`, so it only lists `plugins/*`. But `update-libs.sh --plugin <name>` resolves against the `--include-skipped` list, so `plugin: ai-core` (or `cotg-ndk`) builds fine — and then this `grep -Fx` matches nothing, exits 1, and under `set -euo pipefail` the command substitution aborts the step with no diagnostic. A green 30-minute build reported as a red run.

Secondary effect of the same line: on a full run the skipped-but-must-compile addons are never checked for a release `.cgp`, so the coverage this step claims ("Every discovered addon produced a release .cgp") is narrower than what was actually built.

Fix: build `all` from `--include-skipped` and subtract `skip.txt` only for the missing-`.cgp` assertion, so the `--plugin` resolution and the coverage check share one list.

tools/addons/src/addons/publish.py
```diff
@@ -0,0 +1,72 @@
...
+    return headers
+
+
+def put(client, bucket: str, key: str, path: Path) -> None:
+    # upload_file streams from disk and retries; put_object would read a
+    # 200 MB addon entirely into memory
+    client.upload_file(str(path), bucket, key, ExtraArgs=headers_for(key))
+
+
+def publish(client, bucket: str, objects: list[tuple[str, Path]],
```
jatezzz, line 53: @hal-eisen-adfa **MEDIUM — publish is non-atomic, so a mid-run failure leaves live bytes that contradict the live catalog's `sha256`.**

This uploads ~100 objects sequentially and the catalog last (call site: `tools/addons/src/addons/cli.py:114`). If the run dies after `dl/apk-analyzer.cgp` is replaced but before `dl/voice-alerts.cgp`, the live `v1/catalog.json` is still the previous one: it advertises the **old** sha256/size for the already-replaced `.cgp`, and any client that verifies the checksum — the whole point of publishing it — rejects a download that is actually intact. The reverse holds if the catalog lands and a later object needs a retry.

Nothing detects or reverts this; the next full publish is the only repair.

Options: write content-addressed object keys and have the catalog swap be the only mutation, or publish to a versioned prefix and flip a pointer, or at minimum re-verify the bucket against the catalog at the end of a run and fail loudly.

tools/addons/src/addons/check.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def check_names(root: Path) -> list[str]:
+    problems = []
+    for path in discover.find_addons(root):
+        directory = path.name
+        name = model.display_name(directory)
+        addon_slug = model.slug(directory)
+
+        if not model.directory_is_valid(directory):
```
jatezzz, line 29: @hal-eisen-adfa **MEDIUM — the PR check validates the directory name but not the slug shape the catalog schema enforces, so a bad name passes CI and breaks the publish.**

`directory_is_valid` only rejects `_`, spaces, empty parts and bad capitalisation. A directory like `Code.Together` or `Layout-Editor+` passes, and `check_names` then compares `pluginName` against the same permissive `slug`, so the PR goes green.

At publish time `catalog.entry` produces `slug: "code.together"`, which fails `site/catalog.schema.json`'s `^[a-z0-9]+(-[a-z0-9]+)*$` inside `jsonschema.validate` — **after** all 23 Gradle builds have run.

Fix: validate the derived slug against that same pattern here in `check_names`, ideally by importing the pattern from the schema so the two can't drift.

.github/workflows/publish-addons.yml
```diff
@@ -0,0 +1,140 @@
...
+        type: boolean
+        default: true
+
+# This job holds the R2 write credentials and runs Gradle builds for every
+# addon directory, so its token gets the least it can do its work with.
+permissions:
+  contents: read
+
+jobs:
+  publish:
```
jatezzz, line 20: @hal-eisen-adfa **MEDIUM — no `concurrency` group on a job that writes shared, mutable live keys.**

Two dispatches (or a `--only` live publish while someone runs a full non-staging publish) interleave `put` calls to the same `dl/`, `p/`, `src/` and `v1/catalog.json` keys. The last catalog written wins, and it can describe a different set of bytes than the ones actually sitting in the bucket — the same checksum-mismatch state as the `publish.py` atomicity comment, but reachable with no failure at all.

```yaml
concurrency:
  group: publish-addons-${{ inputs.staging }}
  cancel-in-progress: false
```

Gating the live path behind a GitHub Environment would also work and adds an approval step for non-staging publishes.

tools/addons/src/addons/publish.py
```diff
@@ -0,0 +1,72 @@
...
+    digest = hashlib.sha256(path.read_bytes()).hexdigest()[:8]
+    stem, _, suffix = path.name.rpartition(".")
+    return f"{stem}.{digest}.{suffix}"
+
+
+def headers_for(key: str) -> dict:
+    suffix = Path(key).suffix
+    headers = {
+        "ContentType": CONTENT_TYPES.get(suffix, "application/octet-stream"),
+        "CacheControl": IMMUTABLE if Path(key).parent.name == "assets" else SHORT,
```
jatezzz, line 40: @hal-eisen-adfa **LOW — the content-hash fix is applied to site assets but not to the two other Cloudflare-cached classes, and the tarball case is checksum-visible.**

`headers_for` grants `IMMUTABLE` only to `assets/`, and `hashed_name` is applied only to `styles.css`, `app.js`, `adfa-logo.svg`. But `p/<slug>.png` and `src/<slug>-src.tar.gz` are also default-cached extensions, so by this PR's own measurement (design §12.4: `.png` came back with `max-age=14400` regardless of the origin header) a browser can hold them for four hours with `max-age=60` set.

The design accepts this because "a tarball's checksum is published in the catalog" — but that's precisely the failure mode: a user who fetched a tarball within the last four hours gets stale bytes whose sha256 does **not** match the fresh catalog, i.e. an apparent integrity failure on good content.

Hashing those two key classes the way the site assets already are costs nothing extra.

site/app.js
```diff
@@ -0,0 +1,162 @@
...
+    node.querySelector('[data-slot="page"]').href = safeUrl(addon.pageUrl);
+    // the tarball is the source deliverable (R39); sourceUrl stays in the
+    // catalog as provenance for consumers that want the repository
+    const src = node.querySelector('[data-slot="source"]');
+    src.href = safeUrl(addon.sourceTarball.url);
+    src.title = `Source tarball, ${size(addon.sourceTarball.size)}`;
+    cards.append(node);
+  }
+  renderActive();
+  status.textContent = shown.length ? ""
```
jatezzz, line 142: @hal-eisen-adfa **LOW — a successfully-loaded but empty catalog leaves the page saying "Loading…" forever.**

If `catalog.json` fetches fine with `addons: []` (a `--only` publish of a since-removed addon, or a bad staging run), `shown.length` is 0 **and** `state.addons.length` is 0, so this ternary falls through to `"Loading\u2026"` and the user stares at a spinner message that will never resolve.

Track an explicit `loaded` flag rather than inferring load state from array length:

```js
status.textContent = shown.length ? ""
  : !state.loaded ? "Loading\u2026"
  : state.addons.length ? "No addon matches these filters." : "No add-ons yet.";
```

tools/addons/src/addons/catalog.py
```diff
@@ -0,0 +1,76 @@
...
+from pathlib import Path
+
+import jsonschema
+
+from addons import discover, model
+
+BASE = "https://addons.appdevforall.org"
+REPO = "https://github.com/appdevforall/plugin-examples"
+# overridden per run so a staging publish links the ref it was cut from
+REF = os.environ.get("ADDONS_SOURCE_REF", "main")
```
jatezzz, line 15: @hal-eisen-adfa **LOW — `sourceUrl` is pinned to `github.ref_name`, so a live publish from a branch bakes a disappearing URL into the catalog.**

`ADDONS_SOURCE_REF: ${{ github.ref_name }}` is whatever branch the dispatch ran from. A live (`staging: false`) publish from a feature branch writes `.../tree/<branch>/plugins/Foo` into every entry; this repo squash-merges and deletes branches, so every "repository" provenance link in the live catalog 404s until the next publish.

Consider forcing `main` when the key prefix is empty (i.e. a live publish), or resolving to the commit SHA, which is durable for both staging and live.

CLAUDE.md
```diff
@@ -111,10 +111,10 @@ If device verification isn't possible in-session, say so explicitly rather than
...
+1. Copy `plugins/Random-XKCD/` — it's the canonical starting template (small but complete, includes the in-IDE help HTML pattern that submissions are expected to follow).
 2. Update `settings.gradle.kts` `rootProject.name`, `build.gradle.kts` `pluginBuilder { pluginName }` + `android { namespace, applicationId }`, and `src/main/AndroidManifest.xml` (`plugin.id`, `plugin.name`, `plugin.main_class`).
 3. Add a row to the README's Examples table.
-4. If your plugin should ship via the website, add it to the `MAP` array in `.github/workflows/update-libs.yml` so the filename mapping picks it up. (`.github/workflows/build-plugins.yml` needs no change — it auto-discovers plugins by scanning `*/build.gradle.kts` for the plugin-builder Gradle plugin.)
+4. Nothing else. Publishing is automatic: `addons discover` finds any directory whose `build.gradle.kts` applies the plugin-builder, and the name, filenames, and URLs all derive from the directory name. The `MAP` array this file used to describe was deleted in #66 and no longer exists.
```
jatezzz: @hal-eisen-adfa **LOW — "Nothing else" is no longer true now that `addons check` gates every PR.**

`check-toolchain.yml` newly runs `addons check` on `pull_request`, which hard-requires, per addon:

- `addon.json` valid against `addon.schema.json`
- `src/main/assets/icon_day.png` **and** `icon_night.png`
- a `<slug>.html` whose `<title>` is exactly the display name and whose `<h1>` contains it
- no `<script>` or inline handler in that page
- `pluginName == slug`, `plugin.name == display_name`, non-empty `plugin.id`
- `plugin.min_ide_version` matching `^\d{2}\.\d{2}$`

A contributor who follows steps 1–4 as written adds a plugin and gets a red PR with no idea why. Suggest replacing "Nothing else" with this list (or a summary) plus a pointer to `docs/plugin-naming-standards.md`.

.github/workflows/update-libs.yml
Daniel-ADFA, line 100: This glob drops the seven root-level addons from GitHub releases.

`scripts/update-libs.sh:145` still builds them via `discover --include-skipped` (ai-agent-gemini, ai-agent-local, ai-agent-mcp, ai-agent-openai, ai-core, ai-literacy-course, cotg-ndk all remain at the repo root at this PR's head), but `plugins/*/build/plugin/*.cgp` no longer matches their output. `fail_on_unmatched_files` stays quiet because `plugins/*` matches something.

The design doc (line 55) says this step is "Untouched. Releases keep shipping module-named .cgp files as they do now." The glob change contradicts that.

Either restore `*/build/plugin/*.cgp` alongside the new pattern, or stop building those modules in the libs refresh and update the doc to say so.

tools/addons/src/addons/publish.py
Daniel-ADFA, line 40: `src/<slug>-src.tar.gz` and `p/<slug>.png` get `max-age=60` here, but the zone overrides that header for Cloudflare's default-cached types. The design doc's own measurement (line 524) shows `.png` served at `max-age=14400` regardless of what we send, and `.gz` is on the same default-cached list (the doc's table at line 500 marks it 60s but did not measure it).

Both keys are fixed across republishes (`cli.py:105-110`), so after a source change `v1/catalog.json` advertises the new `sourceTarball.sha256` within 60s while the tarball can serve old bytes for up to four hours. Anyone verifying the checksum gets a mismatch. Changed icons show stale for the same window. `.cgp` is fine because it is not a default-cached type.

Cheapest fix: hash these keys the same way `assets/` already does (`src/<slug>-src.<8hex>.tar.gz`, `p/<slug>.<8hex>.png`) and reference the hashed name from the catalog and page template. Alternative: a Cloudflare Cache Rule for `src/*` and `p/*.png` that respects origin headers.

.github/workflows/update-libs.yml
hal-eisen-adfa: That was very deliberate. I know that you are working on the AI plugins and I didn't want to touch them until your work settled down. The remaining plugins will be migrated later.

tools/addons/src/addons/publish.py
hal-eisen-adfa: This would be over-engineering what should be a simple process.

.github/workflows/publish-addons.yml
hal-eisen-adfa: More overengineering. We're a small team. This won't happen.

tools/addons/src/addons/publish.py
hal-eisen-adfa: A deliberate choice

site/app.js
hal-eisen-adfa: Overengineering. Won't ever happen

.github/workflows/update-libs.yml
hal-eisen-adfa: Deliberate. I'm avoiding conflicts with John's AI work.

tools/addons/src/addons/publish.py
hal-eisen-adfa: By design.

.github/workflows/build-plugins.yml
hal-eisen-adfa: Agreed

tools/addons/src/addons/check.py
hal-eisen-adfa: Ok, fair

tools/addons/src/addons/catalog.py
hal-eisen-adfa: Not important at this time. Might come back to this down the road.

CLAUDE.md
hal-eisen-adfa: Right idea, wrong solution. But I will fix it.

jatezzz (approved):

PR #88, opened 2026-09-07
ADFA-5395 | Adopt build provenance in AI plugins and release pipeline

jatezzz:
## Description

Adopted build provenance placeholders across the AI plugins, documented the provenance contract, and integrated automated enforcement in the release pipeline.

* **AI Plugins**: Added `${pluginVcsRevision}` and `${pluginBuildTimestamp}` to `AndroidManifest.xml` for `ai-core`, `ai-agent-gemini`, `ai-agent-local`, `ai-agent-mcp`, `ai-agent-openai`, and `get-ai-models`.
* **Versions**: Bumped versions for `ai-agent-local`, `ai-core`, and `get-ai-models`.
* **CI/Scripts**: Updated `scripts/update-libs.sh` to export a 12-character `PLUGIN_LIBS_REVISION`.
* **Provenance Verification**: Added a `verify_provenance` step in `scripts/update-libs.sh` to fail the build if `cgp-build.properties` is missing or invalid.
* **NDK Installer Fix**: Untracked `ndk-cmake.tar.xz` in `ndk-installer-plugin` and added it to `.gitignore` to prevent persistent `+dirty` builds.
* **NDK Assembly Guard**: Updated the `ndk-installer-plugin` build file to explicitly fail if the NDK asset is missing prior to assembly.
* **Documentation**: Added a comprehensive "Build provenance" section to `CLAUDE.md`.
* **Code Review**: Updated the `plugin-review` skill rubric to check for missing or `+dirty` provenance in artifact submissions.

## Details

https://github.com/user-attachments/assets/5130913e-1270-4a8a-b329-00a1272cd477

image

## Ticket

[ADFA-5395](https://appdevforall.atlassian.net/browse/ADFA-5395)

## Observation

As noted in the new documentation inside `CLAUDE.md`, the `libs_revision` effectively maps to the commit *before* the "chore: update libs" commit. Ensure the `update-libs` action has refreshed `libs/` before building these plugins, or the manifest merger will fail outright.

[ADFA-5395]: https://appdevforall.atlassian.net/browse/ADFA-5395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/update-libs.sh
```diff
@@ -168,6 +181,78 @@ if [ -n "$ONLY_PLUGIN" ]; then
...
+    echo "Error: unzip is required to verify the provenance record inside each built .cgp." >&2
+    exit 1
+fi
+
+# The plugin builder writes assets/cgp-build.properties into every .cgp (see the
+# Build provenance section of CLAUDE.md). Nothing in the Gradle build fails when
+# that record is missing, truncated or unreadable — the archive still assembles
+# either way — so a builder or asset-packaging regression would ship provenance-less
+# artifacts and only be noticed once a crash report could no longer be traced to
+# a commit. Assert the record here: both release workflows drive their builds
```
Daniel-ADFA: IMPORTANT: neither the `PLUGIN_LIBS_REVISION` export nor this assertion reaches the workflow that publishes addons.

`.github/workflows/publish-addons.yml` builds every addon itself - `cd "$dir"`, then `"$gradlew" downloadAssets` and `"$gradlew" assemblePlugin` - and never sources this script. So each `.cgp` uploaded to R2 is built with `PLUGIN_LIBS_REVISION` unset, carries no `libs_revision` key at all, and is never checked against `REQUIRED_PROVENANCE_KEYS`. The artifacts users install from the gallery are exactly the ones the new pairing is missing from. Only `build-plugins.yml` and `update-libs.yml` run through here, and three sentences this PR adds say otherwise: this comment, CLAUDE.md:121 and CLAUDE.md:123.

Export the revision and run `verify_provenance` in `publish-addons.yml` too, or lift both into one script all three workflows call.

plugins/AI-Core/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,16 @@
             android:name="plugin.version"
             android:value="${pluginVersion}" />
 
+        <!-- Build provenance (ADFA-5256): the commit and commit-date this .cgp was
+             built from. Both placeholders are always supplied by the plugin builder. -->
+        <meta-data
+            android:name="plugin.vcs_revision"
```
Daniel-ADFA, line 24: MINOR: adopting the placeholders makes these eleven examples unbuildable on device in CoGo 26.36 and earlier.

CLAUDE.md:107, added by this PR, states the coupling: the on-device builder ships in `plugin-maven-repo.zip` and refreshes only with an app release, and a manifest naming a placeholder the builder does not define fails the merger outright. ADFA-5394 landed on CoGo `main` in `a76744f49` on 2026-09-07, so 26.37 (2026-09-08) is the first release that carries it. A user still on 26.36 who opens AI-Core on device and builds it gets *"requires a placeholder substitution but no value ... is provided"* with nothing tying it to the builder version.

Nothing in the repo fixes this - `plugin.min_ide_version` (26.35 here) gates installing the plugin, not building its source - so say so in the PR body and in CLAUDE.md rather than leaving the floor implicit.

Daniel-ADFA (commented):
Reviewed a7950e9 at effort high. Every finding was re-derived from the diff, from the committed `libs/gradle-plugin.jar`, and from CodeOnTheGo at `origin/main` - none is taken on the strength of the review's prose.

IMPORTANT
- scripts/update-libs.sh:194 - publish-addons.yml bypasses this script, so published .cgp files get neither `libs_revision` nor the assertion

MINOR
- plugins/AI-Core/src/main/AndroidManifest.xml:24 - on-device builds of these eleven examples fail on CoGo 26.36 and earlier
- plugins/NDK-Installer/build.gradle.kts:169 - a name filter unwires the new guard silently, against this repo's own stated convention
- plugins/NDK-Installer/build.gradle.kts:171 - AI-Literacy-Course has the same missing-asset hazard and no guard
- scripts/update-libs.sh:201 - `head -n1` can verify a leftover .cgp rather than this run's

NITPICK - 2 inline. One more not posted: the `command -v unzip` preflight at line 184 runs after `libs/` has already been overwritten at line 113, so a host without unzip exits 1 with the jars replaced and nothing built; it belongs beside the gradlew check at line 85.

Checked and clear, recorded because this PR's own docs name them as hazards: the committed `libs/gradle-plugin.jar` (byte-identical on `main`) sets both `pluginVcsRevision` and `pluginBuildTimestamp` as unconditional manifest placeholders, so the coupling at CLAUDE.md:107 is satisfied for host builds; the builder reads `PLUGIN_LIBS_REVISION` through `project.providers.environmentVariable`, not `System.getenv`, so a warm Gradle daemon cannot silently drop it; `none` and `wall-clock` match the enum ids in `PluginProvenance.kt`, so both warning branches are live; `build/` is ignored at the repo root and per plugin and the resolver scopes `git status` to the plugin directory, so the `+dirty` fix holds; and no third plugin registers `downloadAssets`.

One thing outside the diff worth knowing: no CI builds plugins on a pull request - `check-toolchain.yml` is the only automatic check and both build workflows are `workflow_dispatch`-only. The eleven adopted manifests have not been proven to merge anywhere but your own build.

This repo has no written approve/request-changes rule, so the default applied.

plugins/NDK-Installer/build.gradle.kts
```diff
@@ -164,9 +164,15 @@ val downloadAssets by tasks.registering {
...
 
-
-
-
-
-
-
+// The archive is never committed, so fail loudly instead of packaging an NDK-less .cgp.
+val ndkArchiveFile = project.file("src/main/assets/ndk-cmake.tar.xz")
+tasks.matching { it.name.startsWith("merge") && it.name.endsWith("Assets") }.configureEach {
```
Daniel-ADFA: MINOR: a name filter unwires this guard silently, which is the failure mode this repo already wrote down.

`plugins/AI-Agent-Local/build.gradle.kts:108` states the convention in as many words: *"tasks.named fails loudly if that name changes, where a name filter would silently stop wiring."* If AGP ever renames `mergeReleaseAssets`, `tasks.matching { ... }` matches nothing, raises no error, and NDK-Installer quietly returns to packaging an NDK-less `.cgp` - the exact regression this hunk exists to prevent. Not reachable on the current AGP, which is why it is minor rather than a bug today.

`tasks.named("mergeReleaseAssets")`, or `tasks.withType<MergeSourceSetFolders>()`, fails at configuration time instead.

plugins/NDK-Installer/build.gradle.kts
```diff
@@ -164,9 +164,15 @@ val downloadAssets by tasks.registering {
...
+    doFirst {
+        if (!ndkArchiveFile.isFile) {
```
Daniel-ADFA, line 174: MINOR: the other `downloadAssets` plugin has the same hazard and no guard.

`plugins/AI-Literacy-Course/build.gradle.kts` fetches `ai-literacy-course.zip` and `pdfjs.zip` onto gitignored paths with no equivalent check, so a build that skips `downloadAssets` still packages a course `.cgp` with no lesson videos and no PDF viewer - the silent failure CLAUDE.md:155 describes, and the one the published source tarball invites, since the BUILDING.md that `tools/addons/tarball.py` generates tells a reader to run `./gradlew assemblePlugin` and nothing else. No CI path reaches it today: `update-libs.sh` and `publish-addons.yml` both run `downloadAssets` first when `build.gradle.kts` mentions it.

Copy this guard into AI-Literacy-Course, keyed on its two zips.

scripts/update-libs.sh
```diff
@@ -168,6 +181,78 @@ if [ -n "$ONLY_PLUGIN" ]; then
...
+# through this script, so this is the one chokepoint that covers all of them.
+REQUIRED_PROVENANCE_KEYS=(name version variant revision revision_source timestamp timestamp_source)
+
+verify_provenance() {
+    local plugin="$1"
+    local cgp props key value revision revision_source timestamp_source recorded_libs
+    cgp="$(ls "$REPO_ROOT/$plugin"/build/plugin/*.cgp 2>/dev/null | grep -v -- '-debug\.cgp$' | head -n1 || true)"
```
Daniel-ADFA: MINOR: `head -n1` verifies whichever release `.cgp` sorts first, not the one this run produced.

`PluginBuilder.createAssembleTask` copies to `<pluginName>.cgp` and never removes siblings, so a `build/plugin/` that has seen a `pluginBuilder { pluginName }` rename holds two release artifacts and this picks the lexicographically smaller one - possibly a stale build with a different `libs_revision`, which then either fails the run with a misleading message or passes while the fresh artifact goes unchecked. `plugins/AI-Agent-Local/build.gradle.kts:117` already deletes leftovers for this exact reason. Unreachable in CI, where `actions/checkout` starts from an empty `build/`; it bites on a repeat local run.

Delete non-matching artifacts first, or assert exactly one release `.cgp` and fail when there are two.

plugins/AI-Agent-Gemini/src/main/AndroidManifest.xml
```diff
@@ -21,6 +21,16 @@
             android:name="plugin.version"
             android:value="${pluginVersion}" />
 
+        <!-- Build provenance (ADFA-5256): the commit and commit-date this .cgp was
+             built from. Both placeholders are always supplied by the plugin builder. -->
+        <meta-data
+            android:name="plugin.vcs_revision"
```
Daniel-ADFA, line 27: NITPICK: eight of the eleven adopting plugins carry this change with no version bump.

AI-Core, AI-Agent-Local and Get-AI-Models bump `versionCode`/`versionName` for exactly this manifest edit; Gemini, MCP, OpenAI, AI-Code-Suggestions, AI-Literacy-Course, Sketch-to-UI, Speech-to-Text and Vector-Search do not. The catalog's `version` field is read from `versionName` (`tools/addons/model.py:55`), so those eight republish under an unchanged version. I could not find the installer's update rule in CodeOnTheGo, so I cannot say whether that suppresses the update on existing installs.

Bump all eleven or none - the split reads as an oversight rather than a decision.

plugins/NDK-Installer/.gitignore
```diff
@@ -27,3 +27,8 @@ Thumbs.db
 
 # Test outputs
 test-results/
+
+# Fetched at build time by the downloadAssets task — never commit it. A tracked
+# copy here would be overwritten by every download, leaving the plugin directory
+# permanently dirty and stamping `+dirty` into the provenance of every .cgp.
+src/main/assets/ndk-cmake.tar.xz
```
Daniel-ADFA, line 34: NITPICK: NDK-Installer does not adopt the placeholders its own bug motivated.

Its manifest declares neither `plugin.vcs_revision` nor `plugin.build_timestamp`, so the one plugin whose committed placeholder caused the `+dirty` problem is a plugin whose provenance the IDE details dialog cannot show. Nothing is lost from the archive - `assets/cgp-build.properties` is written either way - only the dialog.

Add the two `meta-data` entries here as well, or say in the PR body why this plugin is held back.

Daniel-ADFA (changes requested):
Updated after the fixes at 4592f64: the finding this block was opened for is fixed, and the block now stands on one new item.

Blocking: `.github/workflows/publish-addons.yml:66`. The `libs/`-subject parse falls through to a `::warning::`, so a publish run whose last `libs/` commit carries an ordinary subject - `d09f05a` is one already in this repository's history - ships every gallery `.cgp` with no `libs_revision`, on a green run. That is the same defect this block was opened for, conditional now rather than certain. Fail the step instead of warning, and check that the parsed value looks like a sha.

The other two findings from that round (the 9-versus-12 character `libs_revision` shape, and the CLAUDE.md paragraph this PR falsified) plus one nitpick are yours to take or leave.

Everything from round 1 is verified fixed, by reading and running the code rather than the replies - the detail is in https://github.com/appdevforall/plugin-examples/pull/88#pullrequestreview-5280688774.

scripts/update-libs.sh
jatezzz: @Daniel-ADFA Right — lifted both into `scripts/verify-provenance.sh`, called from `update-libs.sh` and directly from `publish-addons.yml` after each `assemblePlugin`.
That workflow now exports `PLUGIN_LIBS_REVISION` too, resolved from the subject of the commit that last touched `libs/` (`gh api`, so no deep fetch).
CLAUDE.md:121 and :123 rewritten to describe the three callers instead of two.

plugins/AI-Core/src/main/AndroidManifest.xml
jatezzz: @Daniel-ADFA Documented: CLAUDE.md:107 now names 26.37 as the first release carrying ADFA-5394, says the source is unbuildable on device in 26.36 and earlier, and that `plugin.min_ide_version` gates installing rather than building.
Called out in the PR body as well.

plugins/NDK-Installer/build.gradle.kts
jatezzz: @Daniel-ADFA Switched to `tasks.withType<MergeSourceSetFolders>()`, which fails at configuration time and keeps covering both variants.
Comment on the hunk now says why it is matched by type rather than by name.

plugins/NDK-Installer/build.gradle.kts
jatezzz: @Daniel-ADFA Copied the guard into `AI-Literacy-Course`, keyed on `ai-literacy-course.zip` and `pdfjs.zip`, naming whichever is missing.
Same typed lookup as NDK-Installer, so a bare `assemblePlugin` from the source tarball now fails instead of shipping a lessonless course.

scripts/update-libs.sh
jatezzz: @Daniel-ADFA Fixed in the extracted script: it collects every release `.cgp` and fails when there is more than one, listing them, instead of taking the first.
Smoke-tested both paths locally (clean dir passes; a planted second artifact exits 1).

plugins/AI-Agent-Gemini/src/main/AndroidManifest.xml
jatezzz: @Daniel-ADFA Oversight — bumped all of them. The eight go `versionCode` 1 -> 2 and `versionName` 1.0.0 -> 1.1.0, the same minor bump Get-AI-Models took.
NDK-Installer is bumped too, since it now adopts the placeholders as well.

plugins/NDK-Installer/.gitignore
jatezzz: @Daniel-ADFA Added both `meta-data` entries to `plugins/NDK-Installer/src/main/AndroidManifest.xml`, so the details dialog shows its provenance like the rest.
No reason to hold it back — it was just missed.

.github/workflows/publish-addons.yml
```diff
@@ -51,10 +51,26 @@ jobs:
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # This workflow builds the addons itself instead of going through
+          # scripts/update-libs.sh, so nothing here knows which CodeOnTheGo commit
+          # produced libs/. "Update libs from CodeOnTheGo" records it in the subject
+          # of the commit it pushes, which is the last one to touch libs/.
+          subject="$(gh api "repos/${GITHUB_REPOSITORY}/commits?path=libs/&sha=${GITHUB_SHA}&per_page=1" --jq '.[0].commit.message' | head -n1)"
+          case "$subject" in
+            *CodeOnTheGo@*) echo "revision=${subject##*CodeOnTheGo@}" >> "$GITHUB_OUTPUT" ;;
+            *) echo "::warning::Could not read a libs revision from '$subject'; the published .cgp files will record no libs_revision." ;;
```
Daniel-ADFA: IMPORTANT: a `libs/` commit with an ordinary subject costs every published .cgp its `libs_revision`, and only warns.

The parse keys on `CodeOnTheGo@` in the subject of the last commit to touch `libs/`. `d09f05a` - *"Publish addons to Cloudflare R2; stop using GitHub artifact storage (#87)"* - is a commit already in this repo's history that touched `libs/` with no such subject. Run Publish addons in that window, or hit a `gh api` failure, and `steps.libs.outputs.revision` is empty: `PLUGIN_LIBS_REVISION` reaches the build as an empty string, the builder omits the key, `verify-provenance.sh:75` skips its comparison because the variable is unset, and the gallery receives artifacts with no pairing on a green run. The annotation is only readable after the publish. The `*CodeOnTheGo@*` case also accepts arbitrary trailing text, so a hand-written subject records whatever follows the `@`.

Fail the step rather than warn, and check the parsed value against `^[0-9a-f]{7,40}$` - on the release path the pairing is the point.

.github/workflows/publish-addons.yml
```diff
@@ -51,10 +51,26 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # This workflow builds the addons itself instead of going through
+          # scripts/update-libs.sh, so nothing here knows which CodeOnTheGo commit
+          # produced libs/. "Update libs from CodeOnTheGo" records it in the subject
+          # of the commit it pushes, which is the last one to touch libs/.
+          subject="$(gh api "repos/${GITHUB_REPOSITORY}/commits?path=libs/&sha=${GITHUB_SHA}&per_page=1" --jq '.[0].commit.message' | head -n1)"
+          case "$subject" in
+            *CodeOnTheGo@*) echo "revision=${subject##*CodeOnTheGo@}" >> "$GITHUB_OUTPUT" ;;
```
Daniel-ADFA: MINOR: this path records a 9-character `libs_revision` where `update-libs.sh` records 12, for the same commit.

`update-libs.yml:80` writes the subject with `git rev-parse --short`, git's default abbreviation, which on that repo yields 9: `chore: update libs from CodeOnTheGo@a76744f49`. `update-libs.sh:144` deliberately re-resolves at `--short=12`, and its own comment gives the reason - two differently shaped shas are needlessly hard to compare. Each run stays self-consistent, so nothing fails; but the same jars now yield `libs_revision=a76744f49` in a gallery artifact and a 12-character form in the one attached to the GitHub release, so the two cannot be compared by equality - which is what the plugin-review rubric's new "cross-check `libs_revision`" step asks a reviewer to do.

Truncate to 12 here, or widen `update-libs.yml`'s subject to `--short=12`.

CLAUDE.md
```diff
@@ -121,9 +150,9 @@ This is intentional — the `application`-as-library packaging trips those check
...
+**Gotcha: a bare `./gradlew assemblePlugin` does NOT run `downloadAssets` and does not warn when the assets are missing** — it silently packages a broken `.cgp` (e.g. a course with no PDF viewer). When building such a plugin by hand, run `./gradlew downloadAssets` and then `./gradlew assemblePlugin` as two separate invocations (or use the script, which does exactly that). Combining them in one invocation fails: `downloadAssets` declares an output inside `src/main/assets`, which Gradle sees as an undeclared dependency of `mergeReleaseAssets`. Then confirm the expected files exist under `src/main/assets/` (or `unzip -l` the `.cgp`) before trusting it. `ndk-installer` is the exception: its asset merge fails outright when `ndk-cmake.tar.xz` is absent, rather than shipping an NDK-less plugin.
```
Daniel-ADFA: MINOR: this paragraph is now false, and this PR is what made it false.

It warns that a bare `assemblePlugin` "silently packages a broken `.cgp` (e.g. a course with no PDF viewer)" and closes with "`ndk-installer` is the exception: its asset merge fails outright". The course example is exactly the case `plugins/AI-Literacy-Course/build.gradle.kts:262` now prevents, and with both download plugins guarded there is no exception left - the remaining plugins have no downloads to miss. A reader trusting this spends time checking `unzip -l` for a silently broken artifact the build can no longer produce, and a future contributor adding a third download plugin reads that guarding is the odd case rather than the rule.

Say both download plugins fail outright, and keep the silent-packaging warning only as the reason the guards exist.

plugins/Vector-Search/build.gradle.kts
```diff
@@ -16,8 +16,8 @@ android {
         applicationId = "com.itsaky.androidide.plugins.vectorsearch"
         minSdk = 33
         targetSdk = 36
-        versionCode = 1
-        versionName = "1.0.0"
+        versionCode = 2
+        versionName = "1.1.0"
```
Daniel-ADFA, line 20: NITPICK: the PR description still names three version bumps; there are twelve.

"**Versions**: Bumped versions for `ai-agent-local`, `ai-core`, and `get-ai-models`" was true at the first commit and stopped being true when the round-1 nitpick was addressed - Vector-Search is one of the nine it does not mention, and NDK-Installer's adoption is missing from the Plugins line too. QA reads the description to decide which gallery entries to re-test, so it will under-test by nine plugins.

Re-read the body against the diff before merge, as CLAUDE.md's "Verify before you claim" asks.

Daniel-ADFA (commented):
Re-review at 4592f64, effort high. Round 1's seven findings are all fixed; two new ones and a doc drift follow from the fixes themselves.

IMPORTANT
- .github/workflows/publish-addons.yml:66 - a `libs/` commit with an ordinary subject downgrades to a warning and publishes .cgp files with no `libs_revision`

MINOR
- .github/workflows/publish-addons.yml:65 - this path records a 9-character `libs_revision` where `update-libs.sh` records 12, for the same commit
- CLAUDE.md:155 - the asset-download gotcha is now false for `ai-literacy-course`, falsified by this PR

NITPICK - 1 inline

Round 1 re-checked at head by reading and running the code, not the replies. All seven fixed:

- `scripts/update-libs.sh:194` (IMPORTANT, publish path uncovered) - **fixed.** `scripts/verify-provenance.sh` is new, committed `100755`, and called from `update-libs.sh:201` and `publish-addons.yml:102` - inside the loop, before the `cp` into `dist/`, under `set -euo pipefail`. CLAUDE.md now names all three workflows.
- `plugins/AI-Core/.../AndroidManifest.xml:24` (MINOR, on-device builds) - **fixed as documentation**, which is what the comment asked for. Worth adding that the window has narrowed since: 26.37 shipped 2026-09-08 and 26.39 is current, so only installs still on 26.36 or earlier are affected.
- `plugins/NDK-Installer/build.gradle.kts:169` (MINOR, name filter) - **fixed.** Both guards use `tasks.withType<MergeSourceSetFolders>()`; I confirmed that type exists in AGP 8.11.0, the version both plugins pin, so the script resolves rather than silently matching nothing.
- `plugins/NDK-Installer/build.gradle.kts:174` (MINOR, literacy course unguarded) - **fixed.** The guard is copied in, keyed on both zips and naming whichever is missing. Its `downloadAssets` aggregates `downloadCourse` and `downloadPdfjs`, so the message's advice really does fetch both.
- `scripts/update-libs.sh:201` (MINOR, `head -n1`) - **fixed, and I ran it**: no artifact exits 1; two release `.cgp` beside a `-debug.cgp` exits 1 listing both; one artifact with a matching `PLUGIN_LIBS_REVISION` exits 0. `publish-addons.yml:104` still selects with `head -n1`, but verification now fails first, so that line can only ever see one file.
- `plugins/AI-Agent-Gemini/.../AndroidManifest.xml:27` (NITPICK, version bumps) - **fixed.** All twelve plugins bumped, `versionCode` and `versionName` both.
- `plugins/NDK-Installer/.gitignore:34` (NITPICK, own adoption) - **fixed.** Its manifest now declares both placeholders.

Also checked and clear: `verify-provenance.sh`'s empty-array handling is safe under the bash 3.2 macOS ships, which this repo explicitly caters to (ran it); neither new `withType` guard fires during `downloadAssets`; and no third plugin registers a download task, so the guarded pair is still the whole set. One thing I deliberately did not raise: the hard error on two release `.cgp` files can abort `update-libs.sh` mid-refresh in a long-lived local checkout that has seen a `pluginName` rename. Failing on an ambiguous state is the right call, and the message says how to clear it.

This repo has no written approve/request-changes rule, so the default applied. One confirmed IMPORTANT means the verdict is unchanged from round 1 - the standing CHANGES_REQUESTED already records it, so no second review is being submitted.

.github/workflows/publish-addons.yml
jatezzz: @Daniel-ADFA Now fails the step (`publish-addons.yml:72`) instead of warning, so an unparseable subject cannot publish artifacts with no `libs_revision` on a green run.
The parsed value is checked against `^[0-9a-f]{7,40}$`, so a hand-written subject can no longer record arbitrary trailing text.
Smoke-tested the three cases: a real subject parses; `"Publish addons to Cloudflare R2 ... (#87)"` and `CodeOnTheGo@not-a-sha` both exit 1.

.github/workflows/publish-addons.yml
jatezzz: @Daniel-ADFA Fixed at both ends: `update-libs.yml:83` now writes the subject with `rev-parse --short=12`, and `publish-addons.yml:76` truncates with `${revision:0:12}` so a longer sha cannot slip through.
The same jars now yield the same `libs_revision` in a gallery artifact and in the one attached to the release, so the rubric's cross-check compares by equality.

CLAUDE.md
jatezzz: @Daniel-ADFA Rewritten: the paragraph now says both download plugins fail the asset merge outright, and there is no exception left.
The silent-packaging warning stays only as the reason the guards exist, with a note to keep one on any new download plugin.

plugins/Vector-Search/build.gradle.kts
jatezzz: @Daniel-ADFA Re-read the body against the diff and rewrote it: 12 plugins (NDK-Installer included), all 12 version-bumped with the nine `1.0.0` -> `1.1.0` named, the AI-Literacy-Course guard added, and `verify-provenance.sh` described as the one script all three workflows call.
The 26.37 floor is now stated in the body as well, which I had claimed in round 1 without actually writing it there.

.github/workflows/publish-addons.yml
```diff
@@ -51,10 +51,35 @@ jobs:
...
+          # scripts/update-libs.sh, so nothing here knows which CodeOnTheGo commit
+          # produced libs/. "Update libs from CodeOnTheGo" records it in the subject
+          # of the commit it pushes, which is the last one to touch libs/.
+          # These are the artifacts the gallery serves, so an unreadable subject fails
+          # the run: a warning is only legible after the publish, and an empty export
+          # makes the builder omit the key and verify-provenance.sh skip its check, so
+          # the pairing would be lost on a green run. The subject is hand-writable, so
+          # the sha is validated rather than taken on trust, and truncated to the 12
+          # characters update-libs.sh records for the same commit so the two compare equal.
+          subject="$(gh api "repos/${GITHUB_REPOSITORY}/commits?path=libs/&sha=${GITHUB_SHA}&per_page=1" --jq '.[0].commit.message' | head -n1)"
```
Daniel-ADFA: IMPORTANT: `per_page=1` makes any ordinary commit that touches `libs/` block the gallery publish entirely.

My round-2 comment asked for the hard failure and did not think this part through, so this is half mine. Looking only at the most recent `libs/` commit means a PR that touches `libs/` without the marker takes the whole release path down: three of the last fourteen such commits are exactly that - `d09f05a` (#87), `a2eae9f` ("add missing jar to libs", #40), `5cd41f2` (ADFA-2298). `d09f05a` is the second most recent, so this would have been live for that whole window. Publish addons then fails before it builds anything, and the only unblock is dispatching Update libs - which is manual, has no schedule, and refreshes the jars from CoGo `stage` as a side effect, so a maintainer who wanted to publish one addon gets a toolchain bump instead.

Raise `per_page`, scan back for the most recent subject that does name a sha, and fail only when none of them does.

.github/workflows/publish-addons.yml
```diff
@@ -51,10 +51,35 @@ jobs:
...
+          revision="${subject##*CodeOnTheGo@}"
+          if [[ "$subject" != *CodeOnTheGo@* || ! "$revision" =~ ^[0-9a-f]{7,40}$ ]]; then
+            echo "Could not read a libs revision from '$subject'." >&2
+            echo "The last commit to touch libs/ must name it as 'CodeOnTheGo@<sha>'." >&2
+            exit 1
+          fi
+          echo "revision=${revision:0:12}" >> "$GITHUB_OUTPUT"
```
Daniel-ADFA: MINOR: truncation cannot lengthen, so the equality this claims is not true of `main` today.

`055e6dc`, the current last `libs/` commit, still carries the 9-character subject the old `rev-parse --short` wrote: `CodeOnTheGo@a76744f49`. `${revision:0:12}` leaves that at 9, so a publish run now records `libs_revision=a76744f49` while the release artifacts built from those same jars through `update-libs.sh` carry the 12-character form. The `--short=12` change fixes it from the next Update libs run onward, not retroactively - but line 68, CLAUDE.md:121 and the PR body all state the equality as current, and the plugin-review rubric asks a reviewer to cross-check the two.

Say it holds from the next refresh, or compare by prefix rather than equality.

CLAUDE.md
```diff
@@ -88,11 +88,40 @@ A plugin is an Android *application* module (despite installing as a library) wi
...
+```xml
+<meta-data android:name="plugin.vcs_revision"    android:value="${pluginVcsRevision}" />
+<meta-data android:name="plugin.build_timestamp" android:value="${pluginBuildTimestamp}" />
+```
+
+**This is a hard build-time coupling to `libs/gradle-plugin.jar`.** A manifest that references a placeholder the builder does not define fails the manifest merger outright (*"requires a placeholder substitution but no value ... is provided"*), and all plugins resolve the builder from the single committed jar. So a manifest may only adopt these **after** the builder change is merged in CoGo and the **Update libs from CodeOnTheGo** Action has refreshed `libs/`. Never the other way round. The same coupling hits on-device builders, whose builder jar ships in `plugin-maven-repo.zip` and refreshes only with a CoGo **app release** — a plugin referencing these cannot be built inside an older CoGo at all. The builder change (ADFA-5394) first ships in **26.37**, so every adopting plugin's source is unbuildable on device in 26.36 and earlier, with only the merger's placeholder error to go on. `plugin.min_ide_version` does not express this: it gates installing the `.cgp`, not building its source.
```
Daniel-ADFA, line 107: MINOR: none of this coupling reaches the people who actually hit it - the tarball readers.

`tools/addons/src/addons/tarball.py:20` generates the `BUILDING.md` shipped inside every published source archive, and it says "Open it in Code on the Go" and `./gradlew assemblePlugin`, nothing else. For the twelve adopters on 26.36 or earlier that now ends at the merger's placeholder error with nothing naming the builder version; for NDK-Installer and AI-Literacy-Course it ends at the new asset guard, which at least names `downloadAssets`. This file and the PR body both explain it, and neither ships in the tarball.

Add two lines to `NOTES`: the 26.37 floor for on-device builds, and `downloadAssets` before `assemblePlugin` for the two download addons.

Daniel-ADFA (commented):
Re-review at 30ec9d3, effort high. Round 2's four findings are all fixed. The first one's fix went further than it needed to and introduced a new problem, which is on my comment as much as on the change - the detail is inline.

IMPORTANT
- .github/workflows/publish-addons.yml:69 - requiring the *last* `libs/` commit to carry the marker makes ordinary repo activity block the only gallery release path

MINOR
- .github/workflows/publish-addons.yml:76 - `${revision:0:12}` cannot lengthen a 9-character sha, so "the two compare equal" is not true of `main` today
- CLAUDE.md:107 - the published source tarballs carry none of this, and `./gradlew assemblePlugin` is now wrong in them for twelve addons

Round 2 re-checked at head, by reading and running the code:

- `publish-addons.yml:66` (IMPORTANT, warn-only) - **fixed.** The step now `exit 1`s. I ran the new parse against six subjects: a 13-character sha truncates to 12; `"Publish addons to Cloudflare R2 ... (#87)"`, `CodeOnTheGo@abc1234 (#99)` and `CodeOnTheGo@not-a-sha` all reject. The trailing-text hole is closed.
- `publish-addons.yml:65` (MINOR, 9 versus 12) - **fixed at both ends**, `update-libs.yml:80` now writes `--short=12` and this step truncates. See the inline comment for what that does and does not cover yet.
- `CLAUDE.md:155` (MINOR, stale gotcha) - **fixed.** It now says both download plugins fail outright, keeps the silent-packaging warning as the reason the guards exist, and adds the rule for the next download plugin.
- `plugins/Vector-Search/build.gradle.kts:20` (NITPICK, stale PR body) - **fixed.** The body now says twelve plugins, names the nine `1.0.0 -> 1.1.0` bumps, adds the asset guards, and states the 26.37 floor.

Round 1 is untouched: `verify-provenance.sh`, `update-libs.sh` and both plugin build files are byte-identical to the blobs I read and executed last round, and the script is still committed `100755`.

Checked and deliberately not raised, so you can see where the line was drawn. `verify-provenance.sh` skips its `libs_revision` comparison when the export is empty, and `revision=unknown` / `+dirty` / `wall-clock` only warn - both are reachable only off-CI, since `actions/checkout` always leaves a `.git` and both download plugins fetch onto ignored paths, and both are the deliberate off-CI tolerance the script documents. `Build plugin artifacts` with a `codeonthego_ref` cut before 2026-09-07 now fails these twelve at the merger, which is the same coupling CLAUDE.md:107 already states rather than a new defect. The stale-sibling `.cgp` abort I declined last round still stands as the right call.

This repo has no written approve/request-changes rule, so the default applied. One confirmed IMPORTANT, so the standing CHANGES_REQUESTED still reflects the verdict and no second review is being submitted.

.github/workflows/publish-addons.yml
jatezzz: @Daniel-ADFA Right — `per_page=50` now, scanning back for the newest `libs/` subject that names a sha and failing only when none of them does, so an ordinary `libs/` commit no longer takes the release path down.
Smoke-tested against real history with `#87` on top (resolves from the commit beneath), and with no valid subject at all (exit 1).

.github/workflows/publish-addons.yml
jatezzz: @Daniel-ADFA Fixed as prefix, not equality: the comment now says the value is *capped* at 12 and that subjects written before `--short=12` carry 9, so `055e6dc` compares by prefix rather than looking like a mismatch.
Same correction in CLAUDE.md:121 and in the rubric line that asks for the cross-check.

CLAUDE.md
jatezzz: @Daniel-ADFA Added to the generated `BUILDING.md`, conditionally per addon rather than to every tarball: the 26.37 floor when the manifest names `${pluginVcsRevision}`, and `downloadAssets` before `assemblePlugin` when `build.gradle.kts` defines the task.
Verified by rendering three real addons — Random-XKCD unchanged, NDK-Installer and AI-Literacy-Course carry both notes; `tools/addons` tests still pass.

Daniel-ADFA (approved):
Approving at 30ec9d3. Every finding from rounds 1 and 2 is fixed and verified - by reading the code and running it, not from the replies - and nothing that reaches a user is at risk here: the manifest entries are inert metadata the committed builder already defines, the asset guards only fail builds that would have shipped broken archives, and `verify-provenance.sh` does what it says when run against real fixtures.

Three things from round 3 are open and deliberately not blocking:

- `publish-addons.yml:69` looks only at the last `libs/` commit, so a PR that touches `libs/` without a `CodeOnTheGo@<sha>` subject will block the gallery publish until someone dispatches Update libs. Not armed today - `055e6dc` carries the marker, so a publish right after merge works - but it springs on the next such commit, and three of the last fourteen were exactly that. Worth fixing before it costs someone an afternoon: raise `per_page` and scan back for the most recent subject that names a sha.
- `publish-addons.yml:76` - the 9-versus-12 equality holds from the next Update libs run, not on `main` as it stands.
- The published source tarballs' `BUILDING.md` needs the 26.37 floor and the `downloadAssets` step.

Details are in https://github.com/appdevforall/plugin-examples/pull/88#pullrequestreview-5281264975. Thanks for the three fast rounds.

PR #89, opened 2026-09-07
chore: refuse hand-built libs/ jars in pull requests

hal-eisen-adfa:
## Why

`Publish addons` failed on `d09f05a` with:

```
java.lang.UnsupportedClassVersionError: com/itsaky/androidide/plugins/build/PluginBuilder
has been compiled by a more recent version of the Java Runtime (class file version 65.0),
this version of the Java Runtime only recognizes class file versions up to 61.0
```

PR #87 committed `libs/` jars built on a laptop running JDK 21, which emits class file version 65. Every workflow here runs JDK 17, which reads up to 61. The tree built fine on the machine that produced it, so nothing caught it until publish time.

`Update libs from CodeOnTheGo` (run 34164643060) has already restored a class-61 jar in `055e6dc`. This PR stops it happening again.

## What

- `Check toolchain` — the only workflow that runs automatically on pull requests — now fails any PR that changes a file under `libs/`.
- `CLAUDE.md` records the policy: run `scripts/update-libs.sh` locally to test a CoGo change, then throw the diff away.

`Update libs from CodeOnTheGo` pushes straight to `main` with an admin PAT that bypasses the `main` ruleset, so it never runs this check. The `main` ruleset requires pull requests, so every human path is covered.

## Note

The `main` ruleset has no `required_status_checks` rule, so `Check toolchain` is advisory — a red PR can still be merged. Adding it as a required check would make this guard binding.

https://claude.ai/code/session_01XWXk1HTqbD46eYmnfAMPNv


hal-eisen-adfa:
Claude went wild and opened this without a ticket or me asking for it

PR #90, opened 2026-09-07
TEST: do not merge — proving the libs/ guard fires

hal-eisen-adfa:
Throwaway PR. It carries the guard from #89 plus a scratch file under `libs/`. If `Check toolchain` passes here, the guard does not work. Closing as soon as the check reports.

https://claude.ai/code/session_01XWXk1HTqbD46eYmnfAMPNv


hal-eisen-adfa:
Guard verified: the check failed on `libs/GUARD_TEST.txt` as intended. Closing the throwaway.

PR #96, opened 2026-09-10
ADFA-5491 | Reveal control on every AI credential field

jatezzz:
## Description

Unified the credential reveal controls across all three AI plugins (Gemini, OpenAI, MCP) to use a consistent `TextInputLayout` end icon touch target. Additionally, this PR ensures that if a new API key is rejected during the save process, the previously stored working key is not overwritten. The user is explicitly informed about the rejection in the settings pane, and the chat continues to function using the retained valid key rather than displaying a generic "No reply" error.

## Details

* Implemented `SecretRevealController` to handle token masking, toggle logic, and dynamic `FLAG_SECURE` window management across plugins.
* Added `CredentialFailureLog` to track credential-specific chat failures so they can be surfaced directly in the settings pane.
* `FLAG_SECURE` is now strictly managed; it is set only while a secret is legible and automatically cleared when it is re-masked or the pane leaves the foreground.
* Updated XML layouts to utilize the `TextInputLayout` custom end icon mode.

https://github.com/user-attachments/assets/4f35214c-a70d-4d8b-8321-52cad749b912

## Ticket

[ADFA-5491](https://appdevforall.atlassian.net/browse/ADFA-5491)

## Observation

The custom end icon setup is deliberately handled in code (via `SecretRevealController`) because Material 1.10 ignores `app:endIconDrawable` attributes declared in XML, which would cause the icon to draw blank.

[ADFA-5491]: https://appdevforall.atlassian.net/browse/ADFA-5491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Daniel-ADFA (commented):
Reviewed at head `62c21f4d`. This is careful work — the four-way `Stored` handling, the `commit()`-gated clear, and the deliberate refusal to overwrite a stored key on rejection are all right, and the two new unit tests pin the classification rather than the wiring. The findings below are about reach: the credential-failure log is cleared on one of three success paths, and two of the three panes disagree about what a credential pane has to do.

## Findings

**IMPORTANT**
- `GeminiBackend.kt:421` — refusal cleared only on the streaming success path; `generate()` / `generateWithHistory()` leave it set (same in `OpenAiBackend.kt:418`)
- `GeminiSettingsFragment.kt:258` — key field never sets `isSaveEnabled = false`, so a plaintext key is parcelled into the saved-state Bundle
- `SecretRevealController.kt:41` — `isEndIconCheckable = true` with `END_ICON_CUSTOM`: the checked state never changes, so TalkBack announces the opposite of the field's state (all three copies)
- `OpenAiSettingsFragment.kt:561` — "chat is still using it" is claimed from a presence check the backend's same-origin rule can override

**MINOR**
- `CredentialFailureLog.kt:19` — a 401 in flight when a new key is saved is recorded against the new key
- `CredentialFailureLog.kt:23` — the persisted sentence is frozen in the locale it was produced in
- `GeminiSettingsFragment.kt:386` — the same "still using it" claim for a key the Keystore can no longer decrypt

**NITPICK** — 3 inline, not listed

## How the findings were checked

- `EditText.getFreezesText()` returning a hardcoded `true` was read in `android/widget/EditText.java` (SDK 36 sources), not assumed.
- The end-icon accessibility claim was checked against `material-1.10.0.aar` bytecode: no `performClick` override on `CheckableImageButton`, `setChecked` reached only from the package-private `EndCompoundLayout.refreshIconState`, no public setter on `TextInputLayout`.
- The same disassembly confirms the PR's own premise: `EndCompoundLayout` reads `endIconMode`, `endIconCheckable`, `endIconContentDescription`, `endIconTint`, `endIconMinSize` and `endIconScaleType` from XML but **never** `endIconDrawable`. Moving the drawable into code is correct and the layout comments are accurate.
- `generate()` and `generateWithHistory()` were confirmed live callers via `LlmInferenceServiceImpl.kt:70` and `:108` before the first finding was graded reachable.

## Raised and dropped

Three candidates did not survive checking, recorded so they are not re-raised:

- **Gemini's layout root has no `android:theme="@style/ThemeOverlay.Plugin"` where OpenAI's does.** That looked like the new `TextInputLayout` would inflate unstyled, or fail into `SafePluginLayoutInflater`'s error view. But `PluginTextInputLayout` / `PluginTextInputEditText` set the box and input colours explicitly on the widget, which is precisely the mitigation the style's own comment describes, and `PluginTheme` derives from `Theme.Material3` so the M3 attributes the parent style resolves are present even when layered without forcing. Not a defect on the evidence available.
- **`onPause()` clearing `FLAG_SECURE` before the re-masked frame is drawn.** `Window.clearFlags` goes through `dispatchWindowAttributesChanged` → `ViewRootImpl.setLayoutParams` → `scheduleTraversals()`, so the flag change and the re-mask land in the same traversal. No window where the flag is gone and the plaintext is still on the surface.
- **Whether a library `app:` attribute reaches the host's Material classes at all.** `dialog_mcp_server.xml` states it does not (citing `MaterialButton`'s `app:icon` drawing empty), while `PluginTextInputLayout` depends on four Material `app:` items taking effect. Only one can be true, and it is not decidable from source — worth one device look at whether the Gemini box actually draws in the plugin's outline colour.

## Verification not done

No device run and no build: this review is source, framework sources and Material bytecode only. Per the repo's CLAUDE.md, `assemblePlugin` succeeding is not verification for these plugins, and the accessibility finding in particular wants a TalkBack pass. The repo has no written approve/request-changes rule, so the default applied.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/backend/GeminiBackend.kt
```diff
@@ -408,6 +416,9 @@ class GeminiBackend(
                         callback.onError("Empty response from Gemini API")
 
                     else -> {
+                        // The saved key was accepted, so any recorded refusal describes a key that
+                        // is no longer in use and must stop being reported in settings.
+                        credentialFailures.clear()
```
Daniel-ADFA: IMPORTANT: the recorded refusal is cleared on only one of three success paths, so a working key keeps being accused.

`credentialFailures.record()` is reached from `formatErrorMessage()`, which serves every path, but `clear()` sits only in the streaming branch. `generate()` (line 239) and `generateWithHistory()` (line 466) complete successfully without clearing, and both are live callers via `LlmInferenceServiceImpl.kt:70` and `:108`. So: a 401 records a refusal, the user fixes nothing, a later non-streaming completion succeeds on the same stored key, and the settings pane still reports "A chat request was refused: ..." under the rejected-key icon against a key that demonstrably works. `OpenAiBackend.kt:418` has the identical gap (success at `:207` and `:255`).

Clear from a single place that every success path routes through, not from the stream branch.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
```diff
@@ -224,35 +242,24 @@ class GeminiSettingsFragment : Fragment() {
...
         }
 
-        applyKeyVisibility()
-
-        toggleVisibilityButton.setOnClickListener {
-            isKeyVisible = !isKeyVisible
-            applyKeyVisibility()
+        // The window is flagged secure for exactly as long as the key is legible, which is why the
+        // click is owned here rather than left to endIconMode="password_toggle".
+        val reveal = SecretRevealController(apiKeyBox, apiKeyInput) { legible ->
```
Daniel-ADFA, line 277: IMPORTANT: this pane never disables saved-state on the key field, so a plaintext key is parcelled into the state Bundle.

`EditText.getFreezesText()` is hardcoded `return true` (`android/widget/EditText.java`), so any `TextInputEditText` with an id writes its full text into `TextView.SavedState`. Tap **Edit** — `revealEditMode` puts the decrypted key into the field at line 408 — or just type one, then rotate, change font scale, or press Home: the key crosses into the saved-state Bundle in clear text. `OpenAiSettingsFragment.kt:418` guards exactly this with `apiKeyInput.isSaveEnabled = false` and a comment saying why; neither `main` nor this PR has an equivalent here, and this PR is what replaced the widget.

```suggestion
        apiKeyInput.isSaveEnabled = false

        val reveal = SecretRevealController(apiKeyBox, apiKeyInput) { legible ->
```

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/ui/SecretRevealController.kt
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Take over [box]'s end icon and mask the field.
+     *
+     * The drawable is set here rather than in the layout because Material 1.10 never reads
+     * `app:endIconDrawable`, so a custom end icon declared in XML draws blank.
+     */
+    fun attach() {
+        box.endIconMode = TextInputLayout.END_ICON_CUSTOM
+        // Announced as a toggle rather than a plain button, which is what it is.
+        box.isEndIconCheckable = true
```
Daniel-ADFA: IMPORTANT: with `END_ICON_CUSTOM` the checked state never changes, so TalkBack announces the opposite of the field's state.

Checked against material-1.10.0 bytecode: `CheckableImageButton` has no `performClick` override, so a tap never toggles it; `EndCompoundLayout` calls `setChecked` only from the package-private `refreshIconState`, which the custom delegate does not drive; and `TextInputLayout` exposes no public setter (`getEndIconView` is package-private). Its accessibility delegate reports `setCheckable(true)` / `setChecked(isChecked())`, so a TalkBack user who reveals the key hears "Hide credential, **not checked**" while the secret is legible — the description and the state contradict each other. Same line in all three copies.

The content description already carries the state; set `isEndIconCheckable = false` and drop the comment's toggle claim.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/settings/OpenAiSettingsFragment.kt
```diff
@@ -541,11 +550,20 @@ class OpenAiSettingsFragment : Fragment() {
...
+                    // Nothing is written: a definitive refusal would only resurface mid-chat. Said
+                    // aloud, because a user who is told the key was refused and then sees chat fail
+                    // concludes the attempt destroyed the key they had, and re-buys a credential
+                    // they never lost.
                     ConnectionVerification.Rejected -> {
                         showStatus(
                             verificationText,
-                            getString(R.string.msg_key_rejected),
+                            getString(
+                                if (viewModel.hasStoredApiKey()) {
```
Daniel-ADFA: IMPORTANT: `msg_key_rejected_kept` promises "chat is still using it" on a check that does not establish the key is sent.

`hasStoredApiKey()` (`OpenAiSettingsViewModel.kt:346`) is a raw `KEY_API_KEY` presence read — correct for dressing the key block, which is what its KDoc scopes it to. But `readApiKeyOrBlank()` (`OpenAiBackend.kt:699`) returns `""` whenever `KEY_API_KEY_URL` is not same-origin with the current base URL. Save a key for `api.openai.com`, point the pane at another server, type a key that server rejects: the pane asserts the old key was kept and chat is still using it, while the backend is sending **no** key at all. The user stops troubleshooting a config that is genuinely broken.

Gate the "kept" wording on the same origin test the backend applies.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/errors/CredentialFailureLog.kt
```diff
@@ -0,0 +1,31 @@
...
+ * by the time they open the pane to fix the key — so the one screen that can fix it said nothing
+ * (ADFA-5491). Holds the sentence the backend already produced for the user, never the credential.
+ *
+ * @param prefs this plugin's settings store; re-read on every call, and null before `initialize()`
+ */
+internal class CredentialFailureLog(private val prefs: () -> SharedPreferences?) {
+
+    /** Records [reason] as the current credential problem, replacing any earlier one. */
+    fun record(reason: String) {
+        prefs()?.edit()?.putString(GeminiPreferences.KEY_CREDENTIAL_FAILURE, reason)?.apply()
```
Daniel-ADFA: MINOR: a refusal in flight when a new key is saved is recorded against the new key.

`record()` is called from the network thread in `formatErrorMessage()`, and nothing ties the reason to the key it describes. Send a request, then save a new verified key while it is outstanding: `saveGeminiApiKey` clears the log (`GeminiSettingsViewModel.kt:163`), the old request's 401 lands afterwards and re-records — so the pane accuses a key that has never been tried, which is exactly what the clear at `:160-163` exists to prevent. Narrow today because it needs a save during an outstanding request, and the window is one request's latency.

Stamp the reason with `KEY_API_KEY_TIMESTAMP` at record time and discard one older than the stored key.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/errors/CredentialFailureLog.kt
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    /** The recorded reason, or null when the credential has not been refused since it was set. */
+    fun read(): String? = prefs()
```
Daniel-ADFA: MINOR: the stored value is an already-rendered sentence, so it is frozen in the locale it was produced in.

`record()` is handed `userMessage(failure)` — a resolved string — and `read()` hands it straight back into `msg_key_chat_failure`. Change the device language between the failed request and opening the pane and the banner reads half in the new locale (the frame) and half in the old (the reason). The same applies when the plugin's own wording is corrected in a later release: the stale sentence outlives the strings file. Not reachable without a language change, so nothing currently misreports.

Record the `GeminiFailure` discriminant instead and resolve it at display time, as `isCredentialProblem` already does.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
```diff
@@ -369,10 +376,19 @@ class GeminiSettingsFragment : Fragment() {
...
-                    // Nothing is written: a definitive refusal would only resurface mid-chat.
+                    // Nothing is written: a definitive refusal would only resurface mid-chat. Said
+                    // aloud, because a user who is told the key was refused and then sees chat fail
+                    // concludes the attempt destroyed the key they had, and re-buys a credential
+                    // they never lost.
                     KeyVerification.Rejected -> {
                         showVerification(
-                            getString(R.string.msg_key_rejected),
+                            getString(
+                                if (viewModel.hasStoredGeminiApiKey()) {
```
Daniel-ADFA: MINOR: "chat is still using it" is also claimed for a key the Keystore can no longer decrypt.

`hasStoredGeminiApiKey()` (`GeminiSettingsViewModel.kt:195`) is a raw ciphertext-presence check, deliberately true for `Stored.Unreadable` and `Stored.Unavailable` — this same fragment handles both at lines 429-446 with "type it again" advice. A user on a restored backup or after an OEM Keystore reset who then gets a new key rejected is told the key they had was kept and chat is still using it; chat has nothing usable. Unreachable without Keystore invalidation, so no ordinary run hits it, but it is the OpenAI sibling's defect in a second form.

Ask the store what state the credential is in before promising chat still works.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
```diff
@@ -138,8 +156,8 @@ class GeminiSettingsFragment : Fragment() {
         val verificationText = view.findViewById<TextView>(R.id.gemini_key_verification_text)
 
         // Not on apiKeyInput: long-press there is the paste menu, and a key is pasted.
-        listOf(
-            toggleVisibilityButton, saveButton, editButton, clearButton, verificationText
+        listOf<View>(
+            apiKeyBox, saveButton, editButton, clearButton, verificationText
```
Daniel-ADFA, line 175: NITPICK: swapping the toggle button for `apiKeyBox` costs the key tooltip its only dedicated long-press target.

`toggleVisibilityButton` was an `ImageButton` that took the long-press directly. `TextInputLayout` does not: a long-press over the input is the paste menu (which the comment above deliberately avoids wiring), and the end icon is a clickable `CheckableImageButton` that consumes the touch, so what is left is the box's few dp of padding. The tag is still reachable from `saveButton`, `editButton`, `clearButton` and `verificationText`, so nothing is lost outright — but the reveal control itself is now the one contributed element with no reachable tooltip.

ai-agent-mcp/src/main/kotlin/com/itsaky/androidide/plugins/aiagentmcp/ui/SecretRevealController.kt
```diff
@@ -0,0 +1,86 @@
...
+ * which is what gives it a real touch target wherever the field is shown. Icon, content
+ * description and toggle behaviour are decided here and nowhere else, so this dialog cannot drift
+ * from the other AI plugins' credential fields (ADFA-5491).
+ *
+ * @param box the field's own layout, whose end icon becomes the control
+ * @param field the masked field
+ * @param onLegibleChanged called with true while the secret stands in clear text, so the caller can
+ *   flag its window secure — which window that is depends on the screen, not on this control
+ */
+internal class SecretRevealController(
```
Daniel-ADFA, line 26: NITPICK: three copies of this class, already diverging on arrival.

The Gemini and OpenAI files are byte-identical apart from the package, the import and one word in a comment; this one adds the two lines that strip the hint and box background. The KDoc's promise that behaviour is "decided here and nowhere else, so this pane cannot drift" holds within a plugin but not across them — a fix to the masking logic now needs three edits. Each addon is an independent Gradle build with only the root `libs/` jars shared, so there is no cheap place to put it today; worth a line in the PR body naming the duplication as deliberate rather than leaving the next reader to infer it.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/preferences/OpenAiPreferences.kt
```diff
@@ -64,6 +64,13 @@ internal object OpenAiPreferences {
      */
     const val KEY_REMEMBERED_MODELS_URL = "openai_remembered_models_url"
 
+    /**
+     * Why the last request was refused for credential reasons, or absent. Diagnostics rather than a
+     * setting: a stale reason would accuse a key the user has since replaced, so it is written by
+     * the backend and forgotten as soon as the settings pane has reported it.
```
Daniel-ADFA: NITPICK: this KDoc contradicts the ViewModel it describes.

"forgotten as soon as the settings pane has reported it" says clear-on-read, but `credentialFailure()` deliberately does not clear, and `OpenAiSettingsViewModel.kt:99-105` spells out why: the view is recreated on every rotation and a read that cleared would drop the message on the first one. The Gemini twin's KDoc has no such sentence. A maintainer following this one would add the clear-on-read and reintroduce the bug the ViewModel guards against.

Say what the Gemini twin says: written by the backend, cleared when a new credential is saved or a request goes through.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/backend/GeminiBackend.kt
jatezzz: @Daniel-ADFA Fixed in c21277ba: the clear moved to the one point every request passes — Gemini's `checkResponse()` on any 2xx, and a new `postChat()` in OpenAI that both `requestText()` and `streamOnce()` go through.
Both stream-branch clears are gone, so `generate()` and `generateWithHistory()` now clear it too.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
jatezzz: @Daniel-ADFA Fixed in c21277ba: `apiKeyInput.isSaveEnabled = false`, with the OpenAI pane's reasoning in the comment.
Applied the same one-liner to the MCP dialog's token field, which this PR gave the same reveal control.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/ui/SecretRevealController.kt
jatezzz: @Daniel-ADFA Fixed in c21277ba: `isEndIconCheckable = false` in all three copies, and the comment now says why rather than claiming a toggle — the content description carries the state.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/settings/OpenAiSettingsFragment.kt
jatezzz: @Daniel-ADFA Fixed in c21277ba: the wording is now gated on `viewModel.getApiKeyFor(viewModel.getBaseUrl()) is Stored.Value`, so the backend's own same-origin rule decides it and an undecryptable key fails it too.
A key saved for another origin now gets the plain "wasn't saved" wording.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/errors/CredentialFailureLog.kt
jatezzz: @Daniel-ADFA Fixed in c21277ba, with the stamp taken where the request reads the key rather than at record time — at record time the pref already holds the new key's timestamp, so it would not discriminate.
`readGeminiApiKey()`/`readApiKeyOrBlank()` set `keyInUseStamp`, and `read()` drops a refusal older than the key on disk.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/errors/CredentialFailureLog.kt
jatezzz: @Daniel-ADFA Fixed in c21277ba: a new `CredentialFailure` enum (persisted tag + `@StringRes`) is stored instead of the sentence, and the pane resolves the wording at display time.
`userMessage()`/`OpenAiFailureMessages.of()` resolve the same ids, so transcript and pane can't drift; `isCredentialProblem` now delegates to its exhaustive `of()`, and a new test pins the tags.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
jatezzz: @Daniel-ADFA Fixed in c21277ba: gated on `viewModel.getGeminiApiKey() is Stored.Value`, so `Unreadable` and `Unavailable` both fall back to the plain "wasn't saved" wording instead of promising chat still works.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
jatezzz: @Daniel-ADFA Fixed in c21277ba: added `wireEndIconTooltip()`, which hangs the key tag off `setEndIconOnLongClickListener` so the icon that consumes the touch is the one that answers it.
Wired in all three panes, since the end icon eats the long-press in each.

ai-agent-mcp/src/main/kotlin/com/itsaky/androidide/plugins/aiagentmcp/ui/SecretRevealController.kt
jatezzz: @Daniel-ADFA Agreed, and named as deliberate in each copy's KDoc rather than only in the PR body: independent Gradle builds sharing only the root `libs/` jars, so a masking change is three edits until the host's plugin-api can carry it.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/preferences/OpenAiPreferences.kt
jatezzz: @Daniel-ADFA Fixed in c21277ba: the KDoc now says what the Gemini twin says — written by the backend, cleared when a new credential is saved or a request goes through, explicitly not on read.

Daniel-ADFA (commented):
Re-reviewed at head `c21277ba` against the round posted on `62c21f4d`. All ten threads were resolved; nine of the fixes hold up when read at head, one does not.

## Re-check of the previous round

**Fixed, verified at head**

- IMPORTANT `GeminiBackend.kt:421` — clear moved into `checkResponse` (`:764-773`), which both `streamContents` (`:378`) and `requestText` (`:721`) call, and it fires on any 2xx. `generate()` and `generateWithHistory()` now clear.
- IMPORTANT `GeminiSettingsFragment.kt` — `apiKeyInput.isSaveEnabled = false` present, and the same one-liner added to the MCP token field, which I had not asked for and which needed it.
- IMPORTANT `SecretRevealController.kt` — `isEndIconCheckable = false` in all three copies, comment rewritten to say why. TalkBack no longer gets a checkable node whose state never moves.
- IMPORTANT `OpenAiSettingsFragment.kt:579` — gated on `getApiKeyFor(getBaseUrl()) is Stored.Value`. `getApiKeyFor` (`OpenAiSettingsViewModel.kt:335-341`) applies the same `savedFor != null && !sameOrigin(...)` test the backend applies at `:733-737`, so the two cannot disagree, and `Unreadable`/`Unavailable` fail it too.
- MINOR `GeminiSettingsFragment.kt:402` — same shape, gated on `Stored.Value`; `getGeminiApiKey()` is `suspend` on `ioDispatcher` and the call sites are inside `lifecycleScope.launch`, so no Keystore IPC moved onto the main thread.
- MINOR `CredentialFailureLog.kt` locale — a `CredentialFailure` tag plus `@StringRes` is persisted and the pane resolves at display time. `isCredentialProblem` now delegates to `CredentialFailure.of`, so the record test and the caller's test are the same exhaustive `when` and cannot drift. `ofTag` returning null makes a tag from an older build forget itself rather than crash.
- MINOR (Gemini half of the in-flight race) — `keyInUseStamp` is taken in `readGeminiApiKey()` at `:136`, **before** the `keyCache` short-circuit on the next line, so a cache hit still stamps. That ordering is what makes it work; had it sat below, every refusal would have been recorded with 0 and silently dropped on read.
- NITPICK tooltip — `wireEndIconTooltip` in all three panes via `setEndIconOnLongClickListener`, with the category constant `plugin_$PLUGIN_ID` and the 3-arg overload. MCP wires it before `attach()` sets the icon mode; checked against material-1.10.0 bytecode that `setEndIconMode` clears neither long-click listener (`tearDownDelegate` only tears down the delegate), so the ordering is safe there.
- NITPICK `OpenAiPreferences.kt:69` — KDoc now matches the ViewModel and says explicitly that the read does not clear.
- NITPICK triplication — accepted as documented rather than changed, which is the right call while each addon is a separate Gradle build.

**Not fixed**

- MINOR `CredentialFailureLog.kt:19` (in-flight race) — fixed in Gemini, inert in OpenAI. The thread is marked resolved; the OpenAI half is still live. Details on `OpenAiBackend.kt:791` below.

## New findings

**MINOR**
- `OpenAiBackend.kt:791` — `formatErrorMessage` re-reads the key, resetting `keyInUseStamp` one line before it is recorded
- `OpenAiBackend.kt:673` — the clear runs after the body is read, so a 2xx that aborts or is cancelled does not clear

## How this round was checked

Claims were verified by reading the code at `c21277ba`, not from the replies. Two things that needed more than reading: `setEndIconMode`'s effect on an already-set long-click listener was disassembled from `material-1.10.0.aar`, and the stamp-ordering question in both backends was traced through every `readApiKeyOrBlank()` / `readGeminiApiKey()` call site rather than the one the comment points at — which is how the OpenAI regression showed up.

Still no device run and no build, as last round. The accessibility change in particular wants a TalkBack pass before QA signs it off, and per the repo's CLAUDE.md `assemblePlugin` succeeding is not verification for these plugins. The repo has no written approve/request-changes rule, so the default applied.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/backend/OpenAiBackend.kt
```diff
@@ -623,15 +641,37 @@ class OpenAiBackend(
...
+        onConnected: (HttpURLConnection) -> Unit = {},
+        readResponse: (BufferedReader) -> T,
+    ): T = http.post(
+        url = getBaseUrl() + CHAT_COMPLETIONS_PATH,
+        apiKey = readApiKeyOrBlank(),
+        body = body,
+        sse = sse,
+        onConnected = onConnected,
+        readResponse = readResponse,
+    ).also { credentialFailures.clear() }
```
Daniel-ADFA: MINOR: the clear runs only when the whole response has been read, so a request the server authenticated can still leave the pane accusing the key.

`.also` fires on `http.post` returning normally. A 2xx that then drops mid-stream, or a generation the user cancels — `currentJob.cancel()` closes the connection and the read throws — skips it, though the server answering is the server accepting the key. The stale refusal survives until the next request completes end to end. Narrow, and self-correcting on the next clean reply, but the Gemini side does not have it: `checkResponse` clears on the response code (`GeminiBackend.kt:773`), before anything is read.

Clear once the status line is known rather than after the body.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/backend/OpenAiBackend.kt
```diff
@@ -741,14 +785,17 @@ class OpenAiBackend(
...
-            OpenAiErrorFormatter.classify(
-                error = e,
-                modelName = getModelName(),
-                hasApiKey = readApiKeyOrBlank().isNotBlank(),
-                isOpenAiHost = BaseUrlPolicy.requiresApiKey(baseUrl),
-            )
+        val failure = OpenAiErrorFormatter.classify(
+            error = e,
+            modelName = getModelName(),
+            hasApiKey = readApiKeyOrBlank().isNotBlank(),
```
Daniel-ADFA, line 792: MINOR: this line resets `keyInUseStamp` immediately before the refusal is recorded, so the stamp never discriminates in this plugin.

`readApiKeyOrBlank()` writes `keyInUseStamp` from the current `KEY_API_KEY_TIMESTAMP` on every call (line 732), and `formatErrorMessage` calls it here to answer `hasApiKey` — one line before `record(failure, keyInUseStamp)` at 797. So the stamp recorded is always the timestamp of the key on disk *now*, and `CredentialFailureLog.read()`'s `keyStamp < KEY_API_KEY_TIMESTAMP` test can never be true. The in-flight case the mechanism was built for still reports against a key that was never tried. `GeminiBackend.kt:871` does not read the key in `formatErrorMessage`, which is why the Gemini twin works.

Capture the stamp where the request picked the key up and pass that, rather than re-reading here.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/backend/GeminiBackend.kt
```diff
@@ -112,6 +131,9 @@ class GeminiBackend(
             keyCache = null
             return null
         }
+        // Every request that sends the stored key picks it up here, so this is where the key in
+        // use is stamped; see keyInUseStamp.
+        keyInUseStamp = agentPrefs()?.getLong(GeminiPreferences.KEY_API_KEY_TIMESTAMP, 0L) ?: 0L
```
Daniel-ADFA: MINOR: one backend-wide field cannot hold a per-request stamp, so Gemini has a non-deterministic form of the OpenAI defect.

Correcting what I said in the body above — Gemini avoids the deterministic bug, not the flaw. `keyInUseStamp` is a single `@Volatile` field, and `readGeminiApiKey()` is also reached from `isAvailable()` (`:226`), which the host calls whenever it re-dresses the backend list, and from every concurrent request. So between a request reading the key and its 401 landing, an `isAvailable()` poll or a second request overwrites the stamp with the newly-saved key's timestamp, and the refusal is filed against a key that has never been tried. Narrow: it needs a save plus a second key read inside one request's latency.

Pass the stamp down with the request that read the key rather than holding it on the backend — the same fix the OpenAI comment asks for.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/errors/CredentialFailureLog.kt
```diff
@@ -0,0 +1,58 @@
...
+     * when a new key is saved lands after the save has cleared the log, and would otherwise accuse
+     * a key that has never been tried.
+     */
+    fun read(): CredentialFailure? {
+        val prefs = prefs() ?: return null
+        val tag = prefs.getString(GeminiPreferences.KEY_CREDENTIAL_FAILURE, null)
+            ?.takeIf { it.isNotBlank() }
+            ?: return null
+        val keyStamp = prefs.getLong(GeminiPreferences.KEY_CREDENTIAL_FAILURE_KEY_STAMP, 0L)
+        if (keyStamp < prefs.getLong(GeminiPreferences.KEY_API_KEY_TIMESTAMP, 0L)) return null
```
Daniel-ADFA: MINOR: after Clear, this guard compares against a timestamp that no longer exists, so a late refusal outlives the key it describes.

`clearGeminiApiKey()` removes `KEY_API_KEY_TIMESTAMP` along with the key (`GeminiSettingsViewModel.kt:204`), and `OpenAiSettingsViewModel.kt:357` does the same. A request already in flight then records with its own positive stamp, and this line reads the removed timestamp as `0` — so `keyStamp < 0` is false and the refusal is kept. The pane ends up showing "A chat request was refused: Invalid API key" on a plugin with no key configured at all, until the user saves a new one. Chat in flight, tap Clear, the 401 lands a second later.

Return null when no key is stored, before the stamp comparison.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
```diff
@@ -369,10 +397,25 @@ class GeminiSettingsFragment : Fragment() {
...
+                    // aloud, because a user who is told the key was refused and then sees chat fail
+                    // concludes the attempt destroyed the key they had, and re-buys a credential
+                    // they never lost.
                     KeyVerification.Rejected -> {
+                        // "Kept, and chat is still using it" only for a key chat can actually
+                        // send. A stored key the Keystore will no longer open — a restored backup,
+                        // an OEM Keystore reset — is on disk but unusable, and telling the user it
+                        // is in use stops them fixing the thing that is actually broken.
+                        val stored = viewModel.getGeminiApiKey()
+                        val keptKeyInUse = stored is KeystoreSecretStore.Stored.Value
```
Daniel-ADFA: MINOR: "kept, and chat is still using it" is also shown when the key just rejected is the stored one.

Edit prefills the field with the decrypted stored key (`revealEditMode`, `:428`). A user whose key was revoked at the provider taps Edit, saves it unchanged to retry, and the check refuses it — `keptKeyInUse` is true, because a decryptable key is on disk, so the pane says the key they had was kept and chat is still using it. It is, and it will keep failing on exactly the credential just proven dead. That is the reassurance ADFA-5491 exists to remove, reached by the most natural retry gesture on the screen. `OpenAiSettingsFragment.kt:580` is the same.

Compare the rejected plaintext with the stored one and fall back to the plain wording when they match.

Daniel-ADFA (commented):
Follow-up to the re-check above, at the same head `c21277ba`. A second pass over the new commit turned up three more, all MINOR, all in the credential-failure mechanism this round introduced. Nothing here changes the verdict: still nothing above MINOR, so nothing blocks the merge.

**MINOR**
- `GeminiBackend.kt:136` — a backend-wide `keyInUseStamp` cannot hold a per-request value; `isAvailable()` and concurrent requests overwrite it
- `CredentialFailureLog.kt:47` — after Clear the compared timestamp is gone, so a late refusal is reported against a plugin with no key
- `GeminiSettingsFragment.kt:410` — re-saving the stored key after revocation still gets the "kept, and chat is still using it" wording

## A correction

The re-check body above says Gemini's stamp works and names `formatErrorMessage` not reading the key as the reason. That is right about the deterministic bug and too strong about the mechanism: `isAvailable()` reaches the same key read, so Gemini has a racy version of the same flaw. The `GeminiBackend.kt:136` comment carries the detail. Both plugins want the same fix, which is the one the OpenAI comment already asks for.

## Raised and dropped

- **`onPause()` clearing `FLAG_SECURE` before the re-masked frame is drawn**, raised again this round on the grounds that the flag is cleared "synchronously through WindowManager". It is not: `Window.clearFlags` delegates to `setFlags`, which only mutates the attributes and calls `dispatchWindowAttributesChanged`, and `ViewRootImpl.setLayoutParams` ends in `mWindowAttributesChanged = true; scheduleTraversals()`. The flag reaches WMS at the next traversal, the same one that draws the masked field. There is a sub-frame window between `relayoutWindow` and `performDraw` inside that traversal, but the pause-to-snapshot latency is orders of magnitude larger, so this is not worth a change on the evidence available.
- **`credentialFailure()` being read once at view creation rather than on resume.** Whether it matters depends on how the host mounts a plugin settings pane — a back-stack replace recreates the view and re-reads; only a retained container would strand the banner. Not decidable from this repo.
- **The unconditional `clear()` scheduling a `SharedPreferences.apply()` when nothing is recorded.** Real, and not worth a guard: `apply()` is asynchronous and this is once per request.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/backend/OpenAiBackend.kt
jatezzz: @Daniel-ADFA Fixed: `OpenAiHttpClient.post` gained an `onAccepted` hook fired right after `failIfNotOk()`, and `postChat` clears there instead of in `.also`.
A 2xx now clears the refusal before a byte of the body is read, so a cancelled or mid-stream-dropped reply counts the same as Gemini's `checkResponse`.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/backend/OpenAiBackend.kt
jatezzz: @Daniel-ADFA Fixed: `readApiKeyOrBlank()` no longer writes the stamp, so `formatErrorMessage` cannot reset it one line before recording.
Each request captures `val keyStamp = storedKeyStamp()` at the top of its `scope.launch` and hands it to `formatErrorMessage(e, keyStamp)`.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/backend/GeminiBackend.kt
jatezzz: @Daniel-ADFA Fixed: the `@Volatile keyInUseStamp` field is gone from both backends; the stamp is now a per-request local, read before the key.
An `isAvailable()` poll or a second request can no longer overwrite it, so a refusal is filed against the key that request actually carried.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/errors/CredentialFailureLog.kt
jatezzz: @Daniel-ADFA Fixed: the guard is now `keyStamp != storedStamp` rather than `<`, so Clear removing the timestamp drops the refusal instead of keeping it.
Equal stamps mean nothing was saved or cleared in between; a keyless OpenAI server (both 0) still reports its refusal.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
jatezzz: @Daniel-ADFA Fixed: `keptKeyInUse` now also requires `stored.plain.trim() != apiKey`, in both panes.
The Edit then save-unchanged retry falls back to the plain "wasn't saved" wording.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/settings/OpenAiSettingsFragment.kt
```diff
@@ -383,38 +416,30 @@ class OpenAiSettingsFragment : Fragment() {
...
-        }
-
-        applyKeyVisibility()
-
         // Not endIconMode="password_toggle": the window has to be flagged secure for as long as the
         // key is legible, and the built-in toggle gives no hook for that.
-        apiKeyBox.setEndIconOnClickListener {
-            isKeyVisible = !isKeyVisible
-            applyKeyVisibility()
+        val reveal = SecretRevealController(apiKeyBox, apiKeyInput) { legible ->
```
Daniel-ADFA, line 444: IMPORTANT: saving a revealed key leaves the window flagged secure with nothing secret on screen. The defect is at line 493, just outside this diff.

The comment directly above this line says the window is flagged secure "for exactly as long as the key is legible". Reveal the key and tap Save: on success the field is collapsed by `updateUiState(isEditing = false)`, but `reveal.mask()` is never called, so `isRevealed` stays true and FLAG_SECURE stays set on the host activity. Screenshots, screen recording and casting of the whole IDE come back black until the next `onPause`, with nothing on screen to explain why. `revealEditMode` already does the equivalent on the way in.

Call `reveal.mask()` alongside `updateUiState(isEditing = false)` on the save-success path.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
```diff
@@ -224,35 +257,30 @@ class GeminiSettingsFragment : Fragment() {
...
+        // Not saved, so a recreate cannot park a typed or revealed key in plain text in the state
+        // Bundle; a stored one is read back from the encrypted store instead.
+        apiKeyInput.isSaveEnabled = false
 
-        toggleVisibilityButton.setOnClickListener {
-            isKeyVisible = !isKeyVisible
-            applyKeyVisibility()
+        // The window is flagged secure for exactly as long as the key is legible, which is why the
+        // click is owned here rather than left to endIconMode="password_toggle".
+        val reveal = SecretRevealController(apiKeyBox, apiKeyInput) { legible ->
```
Daniel-ADFA, line 284: IMPORTANT: saving a revealed key leaves the window flagged secure with nothing secret on screen. The defect is at line 333, just outside this diff.

Same as the OpenAI pane: the save-success path calls `updateUiState(isEditing = false)` and collapses the field without calling `reveal.mask()`, so `isRevealed` stays true and FLAG_SECURE stays set on the host activity until the next `onPause`. `revealEditMode` at line 442 already masks on the way in, so the two directions disagree.

Call `reveal.mask()` alongside `updateUiState(isEditing = false)` on the save-success path.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/settings/OpenAiSettingsFragment.kt
```diff
@@ -116,10 +121,23 @@ class OpenAiSettingsFragment : Fragment() {
...
 
+    /**
+     * Re-mask a revealed key on the way out of the foreground.
+     *
+     * Re-masking rather than only clearing [WindowManager.LayoutParams.FLAG_SECURE]: the flag has
+     * to go, since the window outlives this pane and nothing else would clear it, and dropping it
+     * over a legible key is what would let the recents thumbnail keep a copy of it.
+     */
+    override fun onPause() {
+        apiKeyReveal?.mask()
```
Daniel-ADFA, line 132: MINOR: the secure flag is dropped in the same call that re-masks, before the masked frame can be drawn.

`mask()` swaps the transformation method, which only invalidates - the masked text reaches the surface on a later frame - and then synchronously calls `onLegibleChanged(false)`, clearing FLAG_SECURE while the surface still holds the revealed key. Whether the task snapshot lands inside that window is timing-dependent and I could not reproduce a capture, so this is about the ordering rather than a proven leak; it is the method's own comment that claims the thumbnail cannot keep a copy. Same shape at GeminiSettingsFragment.kt:125 and McpSettingsFragment.kt:117.

Clear the flag once the field has drawn, or in `onStop`, rather than inside `mask()`.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/errors/CredentialFailureLog.kt
```diff
@@ -0,0 +1,58 @@
...
+            ?.apply()
+    }
+
+    /**
+     * The recorded failure, or null unless it describes the credential that is on disk now.
+     *
+     * Equal stamps mean nothing was saved or cleared between the refused request reading the key
+     * and now, so a refusal that landed after either is dropped rather than reported.
+     */
+    fun read(): CredentialFailure? {
```
Daniel-ADFA, line 40: MINOR: a recorded refusal outlives the origin it belongs to.

`read()` validates the stored failure against `KEY_API_KEY_TIMESTAMP` alone, while the key itself is origin-scoped - `OpenAiBackend.readApiKeyOrBlank` drops a key whose `KEY_API_KEY_URL` is not the same origin as the current base URL. Point the plugin at a local Ollama without touching the key: the backend correctly sends no key, the stamp still matches, and the pane reports a refused chat request against a server that never saw the credential, until some later request succeeds and calls `clear()`.

Record the origin with the failure, or compare `KEY_API_KEY_URL` here as well.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/settings/OpenAiSettingsFragment.kt
```diff
@@ -383,38 +416,30 @@ class OpenAiSettingsFragment : Fragment() {
                     ).show()
                 }
             }
+            // A request the server refused for credential reasons is reported here too, and named:
+            // this pane is where the key gets fixed, and the transcript that carried the reason has
+            // been left behind by the time the user arrives.
+            viewModel.credentialFailure()?.let { failure ->
```
Daniel-ADFA: MINOR: a refusal recorded while the pane is already open is never shown.

`credentialFailure()` is read once, here, inside the coroutine that runs at view creation. The fragment carries `onPaneResume` precisely because it gets resumed without the view being rebuilt, so a refusal that lands after this point - a chat request still in flight when the user navigates into settings - leaves the banner blank until the view is destroyed and recreated. That is the case the log exists for: the one screen that can fix the key says nothing. Same shape at GeminiSettingsFragment.kt:262.

Re-read `viewModel.credentialFailure()` from `onPaneResume` as well.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/backend/OpenAiBackend.kt
```diff
@@ -677,6 +710,15 @@ class OpenAiBackend(
     }
 
     /** The stored key as a possibly-empty string, for the calls that treat "no key" as valid. */
```
Daniel-ADFA: NITPICK: this KDoc now documents the wrong function.

`storedKeyStamp()` was inserted between this one-line KDoc and `readApiKeyOrBlank()`, which it described. It now sits directly above `storedKeyStamp()`'s own KDoc block and describes neither function.

Move it back to immediately above `readApiKeyOrBlank()`.

Daniel-ADFA (commented):
Reviewed at `cb4cbaa`. Not a stacked PR - based on `main`, nothing stacked on it.

IMPORTANT
- OpenAiSettingsFragment.kt:493 - saving a revealed key leaves FLAG_SECURE set with nothing secret shown
- GeminiSettingsFragment.kt:333 - same defect in the Gemini pane

MINOR
- OpenAiSettingsFragment.kt:132 - secure flag dropped before the re-masked frame is drawn
- CredentialFailureLog.kt:40 - a recorded refusal is not scoped to the key's origin
- OpenAiSettingsFragment.kt:422 - a refusal recorded while the pane is open is never shown

NITPICK - 1 inline, not listed

The two IMPORTANT findings are anchored on the nearest line inside the diff; both true sites are pre-existing lines just outside it, named in the first sentence of each comment.

Checked and found sound, so as not to re-litigate them later: the reveal contract holds in all three plugins (`mask()` on pause, FLAG_SECURE tied to legibility, cleared on `onDestroyView`); MCP correctly flags both the dialog and the activity window, the latter being what the recents thumbnail captures; `isSaveEnabled = false` closes the saved-state Bundle leak; `CredentialFailureLog` persists a classification tag and never the credential; the stamp protocol resolves a refusal-after-replacement in the conservative direction; and all three modules carry the drawables and string ids their controllers reference.

This repo has no written approve/request-changes rule - no REVIEW.md, no CONTRIBUTING.md, no PR template, and CLAUDE.md does not set one - so the default applied: any confirmed IMPORTANT means changes requested.

Daniel-ADFA (changes requested):
Two IMPORTANT findings, both the same defect and both a one-line fix: the save-success path collapses the key field without calling `reveal.mask()`, so `isRevealed` stays true and FLAG_SECURE stays set on the host activity with nothing secret on screen. That breaks the invariant the code states for itself - the window is secure "for exactly as long as the key is legible" - and `revealEditMode` already masks on the way in, so the two directions disagree. See OpenAiSettingsFragment.kt:493 and GeminiSettingsFragment.kt:333.

The three MINOR findings and the nitpick are worth a look but do not block.

The security design here is good and I want to be clear that this is two small gaps rather than a troubled change: the reveal contract holds across all three plugins, MCP flags both the dialog and the activity window, `isSaveEnabled = false` closes the saved-state Bundle leak, and the failure log keeps a classification tag and never the credential.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/settings/OpenAiSettingsFragment.kt
jatezzz: @Daniel-ADFA Fixed: `persistKey`'s success path now calls `reveal.mask()` before `updateUiState(isEditing = false)`, so the flag comes down with the field.
Same one-liner in the Gemini pane.

ai-agent-gemini/src/main/kotlin/com/itsaky/androidide/plugins/aiagentgemini/settings/GeminiSettingsFragment.kt
jatezzz: @Daniel-ADFA Fixed: `reveal.mask()` alongside `updateUiState(isEditing = false)` in `persistKey`, so the save path and `revealEditMode` now agree.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/settings/OpenAiSettingsFragment.kt
jatezzz: @Daniel-ADFA Fixed in all three copies: `apply()` now reports the mask through a new `afterNextDraw {}` — a Choreographer frame callback that posts a message, so it lands after that frame's traversal rather than before the draw.
Two guards came out of the MCP dialog: a detached field reports immediately, since the dismiss path would otherwise never run the post, and the deferred call re-checks `isRevealed` so a re-reveal within the frame cannot clear the flag.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/errors/CredentialFailureLog.kt
jatezzz: @Daniel-ADFA Fixed: `read()` applies the backend's own same-origin test (`KEY_API_KEY_URL` vs `KEY_BASE_URL`) after the stamp check, so a refusal from a server that never saw the key is dropped.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/settings/OpenAiSettingsFragment.kt
jatezzz: @Daniel-ADFA Fixed in both panes: the block moved into a local `showCredentialFailure()`, called from view creation and from `onPaneResume`.
It sits in the `else` of the key-page hint, so the two never fight over the status line.

ai-agent-openai/src/main/kotlin/com/itsaky/androidide/plugins/aiagentopenai/backend/OpenAiBackend.kt
jatezzz: @Daniel-ADFA Deleted rather than moved: it duplicates `readApiKeyOrBlank()`'s own block KDoc directly below it, so moving it back would restore the double KDoc that is already on `main`.

Daniel-ADFA (approved):

PR #97, opened 2026-09-14
ADFA-5583 | Scope chat sessions to the active project

jatezzz:
## Description

This PR ensures that AI chat interactions are securely scoped to the user's active project to prevent conversation bleed across different codebases.

* Introduced a `ProjectKey` generator to create stable SHA-256 digests of the active project root, preventing global namespace collisions.
* Updated `ChatStorageManager` and `ChatSession` to support a `projectKey` property, writing preference keys isolated per project.
* Implemented a one-time migration in `ChatStorageManager` to gracefully transition legacy global chat history into the currently active project's namespace and clean up old global keys.
* Modified `ChatViewModel` to proactively persist chat states (upon creation, modification, deletion, and message sync) using debounced saves in a dedicated, independent IO CoroutineScope instead of relying solely on `onCleared`.
* Updated `ChatFragment` to initialize and re-sync storage securely when the project root changes, prioritizing flushing the outgoing project's data first.

## Details

Logic-related changes primarily involve state management and storage mechanisms. Unit tests have been significantly expanded in `ChatStorageManagerTest`, `ProjectKeyTest`, `ChatViewModelSessionStateTest`, and `ChatViewModelStorageTest` to ensure that:

* Legacy data is properly adopted by the current project.
* Saving/loading chat states routes strictly to the active namespace.
* Rapid sequential state updates are safely captured and persisted on background threads.

### Before

https://github.com/user-attachments/assets/340755f8-b07e-4db2-bf15-8e430ed65ac8

### After

https://github.com/user-attachments/assets/fd7bb24c-e65c-4f1d-840c-e208a556ec42

## Ticket

[ADFA-5583](https://appdevforall.atlassian.net/browse/ADFA-5583)
Parent: [ADFA-5582](https://appdevforall.atlassian.net/browse/ADFA-5582)

## Observation

To ensure smooth operation and prevent data loss during sudden IDE teardowns, `ChatViewModel` now manages storage via an independent `persistScope` coupled with a Mutex block. This ensures that saving states doesn't get abruptly canceled just because the `viewModelScope` is closing during tab switches or project changes.

[ADFA-5583]: https://appdevforall.atlassian.net/browse/ADFA-5583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Daniel-ADFA (approved):

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/managers/ProjectKey.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * The namespace for the project open right now. Safe to call from the main thread; it asks the
+     * host for the open project and does no I/O beyond canonicalizing the path it gets back.
+     *
+     * @return the current project's key, [NO_PROJECT] when no project is open, or null when the
+     *   host could not be asked at all. Null is not "the project changed" — a caller holding a
+     *   binding should keep it rather than swap the user onto an empty history over a hiccup.
+     */
+    fun current(): String? = try {
```
dara-abijo-adfa, line 39: Distinguish between null and "noproject" so namespaces are not switched mistakenly.


```suggestion
    fun current(): String? = try {
    
    fun current(): String? = try {
        val rawRoot = PathGuard.rawProjectRoot() // return null if unresolved
         if (rawRoot == null) null else forRoot(rawRoot)
    } catch (e: Exception) {
        null
    }
```

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/managers/ChatStorageManager.kt
```diff
@@ -2,70 +2,182 @@ package com.itsaky.androidide.plugins.aicore.managers
...
 
     /**
-     * Replaces the stored history with [sessions].
+     * Replaces this project's stored history with [sessions].
      *
-     * @param sessions every session to keep; an empty list clears the history
+     * @param sessions a snapshot of every session to keep, owned by the caller and not being
+     *   mutated elsewhere; an empty list clears the history
      */
     fun saveSessions(sessions: List<ChatSession>) {
```
dara-abijo-adfa: You might want to combine both updates (`saveSessions` and `saveCurrentSessionId`) into a single transaction instead of calling them back-to-back inside `persistMutex` in `ChatViewModel - persistState()`

If the process dies between both calls, `chat_sessions_{KEY}` and `current_session_id_{KEY}` can get out of sync. Also, since the code is being executed in `persistScope`, calling `apply()` on the main thread can cause an ANR during `onStop` or `onDestroy`.

Something like this can help:
```
fun persist(sessions: List<ChatSession>, currentSessionId: String?) {
    val prefs = prefs ?: return
    try {
        prefs.edit()
            .putString(sessionsKey, gson.toJson(sessions))
            .putString(currentSessionIdKey, currentSessionId)
            .commit() // Safe and synchronous on Dispatchers.IO
    } catch (e: Exception) {
        Log.e(TAG, "Could not persist chat state for project $projectKey", e)
    }
}
```

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/managers/ChatStorageManager.kt
```diff
@@ -2,70 +2,182 @@ package com.itsaky.androidide.plugins.aicore.managers
...
+    }
+
+    /**
+     * @param json a stored sessions blob.
+     * @return the sessions it holds, or an empty list when it cannot be read.
+     */
+    private fun parseSessions(json: String): List<ChatSession> {
+        val type = object : TypeToken<List<ChatSession>>() {}.type
+        return try {
+            gson.fromJson<List<ChatSession>?>(json, type)?.filterNotNull() ?: emptyList()
```
dara-abijo-adfa: Sanitize messages to avoid NPEs.
If a partially saved session has a missing or null `messages` field , Gson leaves `messages` as null, despite being a non-nullable type as in `List<ChatMessage>` in `ChatSession`. This can lead to an NPE when calling `firstOrNull()`.


```suggestion
            gson.fromJson<List<ChatSession>?>(json, type)?.filterNotNull() ?: emptyList()
            
            gson.fromJson<List<ChatSession>?>(json, type)
                ?.filterNotNull()
                ?.map { it.copy(messages = it.messages ?: emptyList()) } ?: emptyList()
```

dara-abijo-adfa (changes requested):

jatezzz:
https://github.com/user-attachments/assets/6446f8db-0875-4f2c-b5d3-57c9e74f510d

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/managers/ProjectKey.kt
jatezzz: @dara-abijo-adfa Good catch — added `PathGuard.rawProjectRoot()`, which skips the `DEFAULT_ROOT` fallback, and `current()` now returns null when no source answers.
`projectRoot()` keeps the fallback so containment still fails closed.
`ChatFragment` already keeps its binding on a null, so a host hiccup no longer swaps the user onto an empty history.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/managers/ChatStorageManager.kt
jatezzz: @dara-abijo-adfa Done — the two methods are now one `persist(sessions, currentSessionId)` that writes both keys on a single editor with `commit()`.
It only ever runs on `Dispatchers.IO` via `persistScope`, so nothing blocks the main thread and no write is left queued for `onStop` to flush.
Removed the single-key methods so the two keys cannot be written apart again.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/managers/ChatStorageManager.kt
jatezzz: @dara-abijo-adfa Added — `parseSessions` now maps every session through a `withMessages` helper that substitutes an empty transcript.
It needs `@Suppress("SENSELESS_COMPARISON")` since Kotlin believes the field cannot be null, which is exactly the trap you spotted.
Covered by a new test that loads a blob with no `messages` field.

dara-abijo-adfa (approved):

PR #99, opened 2026-09-15
ADFA-5584 | Restore LLM context from visible transcript

jatezzz:
## Description

This PR fixes an issue where the model would completely forget a conversation after switching sessions, despite the user interface successfully restoring previous messages.

* **What**: Rebuilds the internal LLM context array when restoring or switching chat sessions.

* **How**: Introduced the `rebuildHistoryFrom` function to convert restored `USER` and `AGENT` messages back into LLM roles. It leaves out `SYSTEM` notices and `TOOL` output (the scaffolding the model never sees live), blank turns, and any agent turn that never finished — one Stop cut off (`durationMs == 0`, the marker `finalizeInProgressMessages` leaves) or one process death interrupted (no duration at all) — so the model is never told it finished saying a half sentence. What survives is spent newest-first against two caps: `MAX_RESTORED_HISTORY` (40 **messages**, so one exchange costs two) and `MAX_RESTORED_HISTORY_CHARS` (8,000 characters, which a long chat carrying code blocks hits first). The newest eligible message is always kept, however long it is, so an oversized paste cannot restore an empty history. `switchToSession`, `deleteSession` and `clearMessages` now rebuild that history and cancel any in-flight run first, so a stale run cannot write its reply into the conversation that replaced the one it was asked for — the overflow menu's Clear chat calls `clearMessages`, and `deleteSession` has no UI calling it yet: it is API prepared for the delete gesture in a follow-up, exercised here by tests only. Comprehensive ViewModel tests were added to validate this logic.

* **Why**: To prevent hallucinations and broken user experiences. If the user sees a conversation on screen, they expect the AI to have the context to "continue" or "explain the previous response."

## Details

All unit tests in `ChatViewModelHistoryRestoreTest.kt` pass, verifying both caps (40 messages and 8,000 characters, including the single message that exceeds the whole budget), exclusion of system/tool scaffolding and unfinished agent turns, the model's own reply text over the rendered bubble, project isolation, and legacy migrations.

### Before

https://github.com/user-attachments/assets/ad2e44c1-223f-4f88-a447-e6b7eb144c23

### After

https://github.com/user-attachments/assets/0298b8bf-03e3-4579-a879-e4e8c8a5e8ed

## Ticket

[ADFA-5584](https://appdevforall.atlassian.net/browse/ADFA-5584)

## Observation

The `clearMessages` function was updated to replace current session messages with an empty list prior to persisting, which prevents cleared chats from returning on the next sync.

[ADFA-5584]: https://appdevforall.atlassian.net/browse/ADFA-5584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1615,8 +1621,12 @@ class ChatViewModel(
         stopStateTimer()
         _messages.value = emptyList()
         _history.value = emptyList()
+        // Without this the session keeps its messages and the cleared chat returns on the next sync.
+        replaceCurrentSessionMessages(emptyList())
         forgetRetryPoint()
         setState(AgentState.Idle)
+        // Written now rather than debounced: a clear is deliberate and must survive a force-stop.
+        persistState()
```
itsaky-adfa, line 1628: IMPORTANT: the Clear chat a user can actually reach does not cancel the run, so the invariant this PR establishes breaks on that path.

The defect is at `ChatViewModel.kt:1635-1642` (`createNewSession`), outside this diff. `ChatFragment.kt:314` wires the overflow menu's Clear chat to `createNewSession()`, not to the `clearMessages()` hardened here, and the overflow button is never disabled while a run is in flight. Tap it mid-answer: `createNewSession()` does not bump `generationEpoch`, so `runModelTurn`'s `isStale()` stays false, the stream callbacks keep calling `syncMessageToSession`, and the reply lands in the session the user just created. The run's `finally` (line 1191) then sets `_history.value` to the old conversation plus that reply, so the "cleared" chat opens holding the context it was cleared to drop.

Your `switchToSession` is safe only because `ChatFragment.syncStorageToCurrentProject` calls `stopProcessing` first. Give `createNewSession` the epoch bump and cancel that `clearMessages` already has.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1615,8 +1621,12 @@ class ChatViewModel(
         stopStateTimer()
         _messages.value = emptyList()
         _history.value = emptyList()
+        // Without this the session keeps its messages and the cleared chat returns on the next sync.
+        replaceCurrentSessionMessages(emptyList())
```
itsaky-adfa, line 1624: MINOR: `clearMessages()` and `deleteSession()` have no callers, so neither change reaches a user.

`git grep clearMessages` and `git grep deleteSession` at 5c12aef find only the declarations (`ChatViewModel.kt:1610` and `:1714`) and this PR's own tests. The only Clear-chat UI, `ChatFragment.kt:314`, calls `createNewSession()`, and nothing in the repo deletes a session. That makes the PR body's Observation -- "prevents cleared chats from returning on the next sync" -- true of no shipping path, and QA exercising Clear chat will see no behaviour change and report the fix as not working.

Either point the menu item at `clearMessages()` (which also cancels the run, unlike `createNewSession()`), or say in the PR body that these two are dead API being prepared for a follow-up.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1669,59 @@ class ChatViewModel(
...
+        )
+        // Starting on an ASSISTANT turn is left alone: every backend here flattens the array, and
+        // trimming back to a USER turn would drop one the user can still see.
+        return kept.map {
+            val role = if (it.sender == Sender.USER) {
+                LlmInferenceService.ChatMessage.Role.USER
+            } else {
+                LlmInferenceService.ChatMessage.Role.ASSISTANT
+            }
+            LlmInferenceService.ChatMessage(role, it.text)
```
itsaky-adfa: MINOR: restored assistant turns carry the IDE's rendered bubble text rather than what the model wrote.

A live run puts `AgentLoop.ModelReply.historyText` -- the model's own reply -- into `_history`, while the saved `ChatMessage` keeps `AgentReplyRenderer.render`'s output. For a turn whose tool call failed that output is `actionFailedText` (`AgentReplyRenderer.kt:57`); for an empty respond envelope it is `noResponseText`, and for an unparsed call it is the localized advice string. After a restart the model is therefore told it previously said "The action failed", in the IDE's display language, and can be asked to explain a sentence it never produced.

Reachable, but it degrades context rather than breaking the app, hence MINOR. Keeping `historyText` on `ChatMessage` beside the display text would let the rebuild use the real turn.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -109,6 +109,12 @@ class ChatViewModel(
         /** Max open files named in the prompt's IDE-context block. */
         private const val MAX_CONTEXT_OPEN_FILES = 8
 
+        /**
+         * How many of a restored transcript's turns the model is given back. Bounds what a long
+         * conversation costs a small local model's context window; see [rebuildHistoryFrom].
+         */
+        private const val MAX_RESTORED_HISTORY = 40
```
itsaky-adfa, line 116: NITPICK: this counts messages, but the KDoc above and the PR body both call them "turns".

`rebuildHistoryFrom` filters and then `takeLast(40)` over individual `ChatMessage`s, so one user/agent exchange spends two of the 40. Someone budgeting a small model's context from "the last 40 turns" will plan for roughly twice the window this restores.

Say "messages" on line 113 and in the `@return` on line 1678, or double the constant if 40 exchanges was the intent.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1669,59 @@ class ChatViewModel(
...
+                it.status != MessageStatus.ERROR &&
+                it.status != MessageStatus.LOADING &&
+                // Gemini rejects an empty content part, and AgentLoop never stores a blank turn.
+                it.text.isNotBlank()
+        }
+        // Capped after filtering, so the budget is spent on turns the model actually sees.
+        val kept = eligible.takeLast(MAX_RESTORED_HISTORY)
+        AgentTrace.detail(
+            "RESTORE",
+            "retained=${kept.size} discarded=${messages.size - kept.size} " +
```
itsaky-adfa: NITPICK: `discarded` already includes the capped turns, so the two counters double-count.

`discarded` is `messages.size - kept.size`, everything not retained; `capped` is `eligible.size - kept.size`, a subset of it. A line reading `retained=40 discarded=60 capped=10` invites the reader to add them and conclude 70 messages were dropped when 60 were -- exactly the arithmetic someone does when a restore looks too short.

Report the filter's own loss instead: `discarded=${messages.size - eligible.size}`.

itsaky-adfa (commented):
## ADFA-5584 | Restore LLM context from visible transcript

`rebuildHistoryFrom` is the right shape for this fix and the tests are unusually good. I ran
`./gradlew testDebugUnitTest --tests "*ChatViewModelHistoryRestoreTest*"` against `5c12aef` in a
clean worktree: `BUILD SUCCESSFUL`, `tests="14" failures="0" errors="0"`. The PR body's claim
holds, and the filter, the cap, project isolation and both legacy-migration cases are each
pinned by a test that would fail without the change.

### Findings by severity

**CRITICAL** - none.

**IMPORTANT**
- `ChatViewModel.kt:1629` - the Clear chat a user can reach (`createNewSession`) never cancels
  the run, so the new invariant breaks on that path.

**MINOR**
- `ChatViewModel.kt:1625` - `clearMessages()` and `deleteSession()` have no callers.
- `ChatViewModel.kt:1707` - restored assistant turns carry rendered UI text, not the model's own.

**NITPICK** - 2 inline, not listed here.

### What I checked, and what that rules out

- `switchToSession` is reached only from `loadSessions()` (`ChatViewModel.kt:436`), so only via
  `initializeStorage`. `ChatFragment.syncStorageToCurrentProject` calls `stopProcessing` before
  rebinding, which bumps `generationEpoch`. An in-flight run therefore cannot stomp the history
  you just rebuilt on the path this PR fixes - that is why the IMPORTANT above is about
  `createNewSession` and not about your change.
- `deleteSession`'s `_sessions.value.firstOrNull()` cannot land on another project's session:
  `loadSessions` replaces `_sessions` wholesale per namespace.
- `takeLast` starting mid-exchange on an ASSISTANT turn is fine for these backends, as the
  comment on line 1699 says.

### Notes on this review

- This repo has no written approve/request-changes rule - no `REVIEW.md`, no `CONTRIBUTING.md`,
  no PR template, and `.claude/skills/plugin-review` scores plugin submissions rather than code
  changes - so the reviewer's default severity rule applied.
- No CI workflow runs unit tests (`build-plugins.yml`, `check-toolchain.yml`,
  `publish-addons.yml`, `update-libs.yml` all skip them), so these 14 tests will not guard the
  fix on future pushes. Pre-existing repo state, not a finding against this PR.
- Per `CLAUDE.md`, a green build is not verification here. I did not device-verify: installing
  the plugin and watching a restored conversation actually continue is still outstanding.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1669,59 @@ class ChatViewModel(
...
+     * @return the last [MAX_RESTORED_HISTORY] eligible turns, oldest first.
+     */
+    private fun rebuildHistoryFrom(
+        messages: List<ChatMessage>
+    ): List<LlmInferenceService.ChatMessage> {
+        val eligible = messages.filter {
+            // SYSTEM notices and TOOL output are the scaffolding this rebuild exists to leave out.
+            (it.sender == Sender.USER || it.sender == Sender.AGENT) &&
+                // A failed or half-streamed turn is not something the model said.
+                it.status != MessageStatus.ERROR &&
```
itsaky-adfa: IMPORTANT: both status checks are inert, and the half-streamed turn the comment names is not the one they catch.

Nothing in `ai-core` ever constructs a `ChatMessage` with `MessageStatus.LOADING` -- only `ChatAdapter` reads it -- so that clause can never fire. `MessageStatus.ERROR` is set in exactly three places: `ChatViewModel.kt:591` and `:1206` on `Sender.SYSTEM`, and `:914` on `Sender.TOOL`. The sender check on line 1685 already excludes all three, so that clause can never fire either.

The real interrupted turn slips through. Press Stop mid-reply: `stopProcessing()` calls `finalizeInProgressMessages()` (line 1769), which rewrites the partial agent bubble to `status = COMPLETED, durationMs = 0L` and mirrors it into the session, where it persists. On restore it passes every clause here and reaches the model as a finished ASSISTANT turn, so the model treats its own truncated sentence as a claim it completed.

`durationMs == 0L` is the marker `finalizeInProgressMessages` leaves; filtering on it would honour the comment. Note this also means `givenAFailedOrHalfStreamedAgentTurn_whenRestoring_thenItIsExcludedFromHistory` passes while pinning two states production never produces.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1669,59 @@ class ChatViewModel(
...
+                it.status != MessageStatus.LOADING &&
+                // Gemini rejects an empty content part, and AgentLoop never stores a blank turn.
+                it.text.isNotBlank()
+        }
+        // Capped after filtering, so the budget is spent on turns the model actually sees.
+        val kept = eligible.takeLast(MAX_RESTORED_HISTORY)
```
itsaky-adfa: MINOR: the cap counts messages, so it does not bound what the KDoc says it bounds.

`MAX_RESTORED_HISTORY` is documented as bounding "what a long conversation costs a small local model's context window", but 40 restored turns are user prose and full agent replies, which routinely carry code blocks. `LocalLlmBackend.buildPrompt` (`LocalLlmBackend.kt:664`) concatenates every entry into one ChatML string ahead of the system prompt and the new message, and `reportEffectiveContextSize`'s KDoc notes a prompt can be rejected for exceeding the created context. So a restored long chat on the local backend can push the next send over that limit where the pre-PR empty history could not.

PLAUSIBLE, not confirmed: I did not trace the rejection to a user-visible failure, and `_history` already grows unbounded within a live session, so this is a sooner-reached limit rather than a new one. Graded MINOR on that basis. A character budget alongside the turn cap would make the bound real.

itsaky-adfa (changes requested):
## Verdict: request changes

Submitted on top of the findings review above (#pullrequestreview-5238653130). Two findings
have landed since, both from a second pass over the filter itself, and the first is what moves
this from "comment" to "request changes".

### Severity index, both reviews

**IMPORTANT**
- `ChatViewModel.kt:1687` - the `ERROR`/`LOADING` clauses can never fire, and the half-streamed
  turn they claim to exclude reaches the model as a completed one.
- `ChatViewModel.kt:1629` - the Clear chat a user can reach (`createNewSession`) never cancels
  the run, so the new invariant breaks on that path.

**MINOR**
- `ChatViewModel.kt:1693` - the cap counts messages, not context, so it does not bound what its
  KDoc says it bounds (PLAUSIBLE).
- `ChatViewModel.kt:1625` - `clearMessages()` and `deleteSession()` have no callers.
- `ChatViewModel.kt:1707` - restored assistant turns carry rendered UI text, not the model's own.

**NITPICK** - 2 inline on the earlier review.

### What this blocks on, and what it does not

The blocking item is `:1687` alone. The fix's shape is right, the restore path is the right one,
and all 14 new tests pass at `5c12aef` (`BUILD SUCCESSFUL`, `failures="0"`) - I ran them. But
the filter that decides what the model is told it said does not exclude anything today, and the
one case it was written for, a reply cut off by Stop, is exactly the case that gets through. That
is worth a round before merge, because a truncated sentence restored as a finished turn is the
same class of wrong context this ticket exists to remove.

`:1629` is pre-existing and outside this diff; it is not a reason to hold this PR, and a
follow-up ticket would be a fair answer to it.

Still outstanding either way: no device verification. Per `CLAUDE.md` a green build is not
verification here, and nobody has yet watched a restored conversation actually continue on a
device.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Fixed. Pulled the cancel + epoch bump out of `clearMessages()` into a shared `cancelActiveRun(reason)`, and `createNewSession()` now calls it.
It also drops back to `AgentState.Idle`, so the run it just cancelled leaves no running UI behind.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Pointed the overflow item at `clearMessages()`, so the hardened path is the one QA reaches.
`deleteSession` is still uncalled; I will say so in the PR body as API prepared for the follow-up.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Added `historyText: String?` to `ChatMessage`, written only when it differs from the rendered bubble.
The rebuild now uses `historyText ?: text`, so pre-existing blobs fall back to the old behaviour.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Both KDocs now say messages, with "one exchange spends two" spelled out; the 40 is unchanged.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Now reports `discarded=${messages.size - eligible.size}` — the filter's own loss, so the two counters no longer overlap.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Confirmed — both clauses were dead. The filter is now `durationMs != 0L`, the marker `finalizeInProgressMessages` leaves.
The test is retargeted at that state and renamed `givenATurnStoppedMidStream_whenRestoring_thenItIsExcludedFromHistory`.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Added `MAX_RESTORED_HISTORY_CHARS = 8_000` beside the message cap.
The rebuild walks newest-first and stops at whichever budget hits first; a new test covers the character cut.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1685,66 @@ class ChatViewModel(
...
+                // Gemini rejects an empty content part, and AgentLoop never stores a blank turn.
+                it.text.isNotBlank()
+        }
+        // Newest-first, so both budgets are spent on the turns nearest the next message.
+        val kept = ArrayDeque<LlmInferenceService.ChatMessage>()
+        var chars = 0
+        for (message in eligible.asReversed()) {
+            // What the model wrote, not the bubble: a turn whose tool call failed renders as
+            // "the action failed" in the IDE's language, a sentence the model never produced.
+            val text = message.historyText ?: message.text
```
itsaky-adfa: CRITICAL: on the native tool-calling path `historyText` is stored as the empty string, so every restored Gemini answer comes back as a blank ASSISTANT turn.

`?:` only falls back on null. When Gemini answers with a `respond` functionCall and no text part, `GeminiBackend` takes the `else` branch at :412 and completes with `LlmResponse.success("")`, so `written = ""`. `respond` is declared as a native tool (`promptToolDefinitions`, :732), so this is the ordinary shape of a Gemini answer, not an edge case. `displayText` is then the rendered answer and :1367 stores `historyText = "".takeIf { it != displayText }` = `""` -- non-null. The filter on :1706 passes because it tests `it.text`, the bubble, which is not blank.

Seeded one such message, the restore emits `contents=[[what does this do], []]`. The model is handed an empty turn where its answer was, and the comment on :1705 is the reason that matters: `AgentLoop:179` guards the live path with `isNotBlank()`, this one does not.

```suggestion
            val text = message.historyText?.takeIf { it.isNotBlank() } ?: message.text
```

itsaky-adfa (commented):
Re-review at `f566cb6` (a merge of `main` into the branch). Five of the seven findings from the previous round are genuinely fixed. The sixth holds for the path it names but not for the invariant its comment states. The remaining two each introduced a new defect while fixing the one they were raised for, and one of those is blocking.

## Severity index

**CRITICAL**
- `ChatViewModel.kt:1714` - an empty `historyText` restores a blank ASSISTANT turn, on the ordinary Gemini path

**IMPORTANT**
- `ChatViewModel.kt:1716` - a single message over the character budget restores nothing at all

**MINOR**
- `ChatViewModel.kt:1704` - the zero-duration marker covers the Stop path only
- `ChatViewModel.kt:1674` - `switchToSession` / `deleteSession` rebuild history without cancelling a run
- `ChatViewModel.kt:1738` - `deleteSession` still has no production caller
- `ChatViewModel.kt:122` - the PR description describes code this branch no longer has

**NITPICK** - 1 inline.

## Previous round, re-checked against the code at `f566cb6`

1. **`createNewSession` does not cancel the run** - **fixed.** `cancelActiveRun` (:1636) is shared and `createNewSession` (:1650) calls it, with the epoch bump inside, so `isStale()` now trips. `cancelActiveRun` does omit the `finalizeInProgressMessages` that `stopProcessing` calls; harmless for `clearMessages` (the transcript is wiped immediately after) and unreachable for `createNewSession`, whose only caller is `loadSessions` on empty storage.
2. **`clearMessages` / `deleteSession` have no callers** - **partly fixed.** `ChatFragment:314` now calls `clearMessages()`, so the hardened path is the one the overflow menu reaches. `deleteSession` is still uncalled; see :1738.
3. **Restored assistant turns carry the rendered bubble** - **fixed for the case it named, broken for a bigger one.** `ChatMessage.historyText` is written at :1367 and read at :1714, and `givenATurnRenderedForDisplay_whenRestoring_thenTheModelsOwnTextIsUsed` pins the localized-`actionFailedText` case. But `?:` does not catch the empty string the native `respond` path stores; see :1714.
4. **"turns" where the code counts messages** - **fixed in the code.** The KDoc at :113 and the `@return` at :1694 both say messages. The PR body still says turns; folded into :122.
5. **`discarded` double-counts the capped messages** - **fixed.** :1728 reports `messages.size - eligible.size`, the filter's own loss.
6. **Both status checks are inert** - **fixed for Stop.** `durationMs != 0L` does catch the marker `finalizeInProgressMessages` leaves, and `givenATurnStoppedMidStream_whenRestoring_thenItIsExcludedFromHistory` pins it. The residue is at :1704.
7. **The cap does not bound what the KDoc says it bounds** - **fixed, with a new defect.** `MAX_RESTORED_HISTORY_CHARS` does bound the restore, but the loop that spends it can return nothing; see :1716.

## Verification

`./gradlew testDebugUnitTest --tests '*ChatViewModelHistoryRestoreTest*'` passes at `f566cb6` (16 tests). Both of the top two findings were confirmed by adding a probe test to that class in a scratch worktree:

- :1714 - a session whose AGENT message carries `historyText = ""` restores as `contents=[[what does this do], []]`.
- :1716 - a session of three messages whose newest is 8001 characters restores `history.size=0`, taking the two short ones with it.

No device verification was possible in this session, so nothing here is confirmed against a running IDE.

This repo has no written approve/request-changes rule (`CLAUDE.md` covers plugin structure and the `/plugin-review` gate, not review verdicts), so the default applied: one confirmed CRITICAL and one confirmed IMPORTANT.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1685,66 @@ class ChatViewModel(
...
+            chars += text.length
+            if (kept.size >= MAX_RESTORED_HISTORY || chars > MAX_RESTORED_HISTORY_CHARS) break
```
itsaky-adfa: IMPORTANT: a single message over the character budget restores an empty history, which is the regression this PR exists to remove.

`chars += text.length` runs before the break, so when the newest eligible message is longer than `MAX_RESTORED_HISTORY_CHARS` the loop breaks on its first pass with `kept` still empty, and every older message goes with it. Seeded a session of three messages whose newest is 8001 characters, `rebuildHistoryFrom` returns `history.size=0`: the user sees the whole transcript and the model is told nothing. The trace then reads `retained=0 discarded=0 capped=3`, pointing at the filter rather than the budget.

A pasted stack trace or a reply carrying a large code block reaches this. `givenTurnsOverTheCharacterBudget` uses three 3 KB messages and never does.

Keep the newest eligible message unconditionally, or truncate its text to the remaining budget, before the budget is allowed to cut.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -109,6 +109,18 @@ class ChatViewModel(
...
+         * How many of a restored transcript's messages the model is given back; one exchange
+         * spends two. See [rebuildHistoryFrom].
+         */
+        private const val MAX_RESTORED_HISTORY = 40
+
+        /**
+         * Character budget for the same restore, since 40 messages carrying code blocks would
+         * otherwise push the next send past a small local model's created context.
+         */
+        private const val MAX_RESTORED_HISTORY_CHARS = 8_000
```
itsaky-adfa, line 122: MINOR: the PR description describes code this branch no longer has, and QA builds its test plan from it.

Three claims are stale at `f566cb6`. It says the rebuild filters "`SYSTEM`, `TOOL`, `ERROR`, and `LOADING`" messages, but the ERROR and LOADING clauses were replaced by the `durationMs` check in 43a49ab. It says the transcript is "capped at the last 40 turns"; the KDoc on :113 was corrected to messages last round and the body was not, so the two now disagree on how much context a restore costs. And it never mentions `MAX_RESTORED_HISTORY_CHARS`, the second cap, and the one most likely to surprise someone testing a long chat.

Rewrite the How paragraph around the two caps and the filter as it now stands.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1644,7 +1670,8 @@ class ChatViewModel(
         _currentSessionId.value = sessionId
         // Use immutable snapshot to ensure StateFlow emits on mutations
         _messages.value = session.messages.toList()
-        _history.value = emptyList()
+        // Emptying this is what had the model forget a conversation the user was looking at.
+        _history.value = rebuildHistoryFrom(session.messages)
```
itsaky-adfa, line 1675: MINOR: `switchToSession` rebuilds the history without the cancel that `clearMessages` and `createNewSession` just gained, and `deleteSession` (:1745) is the same.

`cancelActiveRun`'s own KDoc gives the reason: without the epoch bump a stale run keeps writing into whatever conversation replaced the one it was asked for. With a run in flight over this call, the `finally` at :1196 (`if (generationEpoch.get() == epoch) _history.value = history.toList()`) overwrites the history you just rebuilt with the previous session's, and the stream callbacks still pass `isStale()`, so the old run's bubble is appended to the newly selected session.

Unreachable today: the only caller is `loadSessions`, and `ChatFragment:254` calls `stopProcessing` before `initializeStorage`. Nothing in the ViewModel enforces that, so the guard belongs here too.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1685,66 @@ class ChatViewModel(
...
+     */
+    private fun rebuildHistoryFrom(
+        messages: List<ChatMessage>
+    ): List<LlmInferenceService.ChatMessage> {
+        val eligible = messages.filter {
+            // SYSTEM notices and TOOL output are the scaffolding this rebuild exists to leave out.
+            (it.sender == Sender.USER || it.sender == Sender.AGENT) &&
+                // Zero is the marker finalizeInProgressMessages leaves on a turn Stop cut mid
+                // sentence, which the model must not be told it finished saying.
+                it.durationMs != 0L &&
```
itsaky-adfa: MINOR: the marker covers Stop only, so the invariant this comment states does not hold for a turn that process death cut.

`onToken` (:1300) writes the partial bubble through `syncMessageToSession`, which schedules a debounced persist, and that bubble carries `durationMs == null`, the data-class default. Only `finalizeInProgressMessages` stamps `0L`, and only `stopProcessing` calls it; `cancelActiveRun` (:1636) does not, and a kill calls neither. Since `null != 0L`, such a bubble restores as a finished ASSISTANT turn -- which this PR's own first test shows a null-duration AGENT message doing.

MINOR rather than IMPORTANT because I could not confirm on a device that the 1s debounce fires mid-stream rather than after it. `onError` (:1383) drops the bubble, so the error path is already clean.

Scope the check to the sender: `it.sender == Sender.USER || (it.durationMs ?: 0L) > 0L`.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1685,66 @@ class ChatViewModel(
...
+                "capped=${eligible.size - kept.size}"
+        )
+        return kept.toList()
+    }
+
     /**
      * Delete a chat session.
      */
     fun deleteSession(sessionId: String) {
         _sessions.value = _sessions.value.filter { it.id != sessionId }
```
itsaky-adfa, line 1745: MINOR: `deleteSession` still has no production caller, and the PR body does not say so as you said it would.

At `f566cb6`, `git grep deleteSession -- ai-core/src/main` finds only the declaration: no fragment, menu item or adapter reaches it, and the only thing exercising it is this PR's own tests. Your reply last round was "deleteSession is still uncalled; I will say so in the PR body as API prepared for the follow-up", and the body still reads "Updates `switchToSession`, `deleteSession`, and `clearMessages`", which reads as three shipping paths. QA will look for a delete gesture and find none.

Add that sentence to the description, or hold `deleteSession` back until the UI that calls it lands.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1685,66 @@ class ChatViewModel(
...
         if (_currentSessionId.value == sessionId) {
             val remaining = _sessions.value.firstOrNull()
+            val messages = remaining?.messages ?: emptyList()
             _currentSessionId.value = remaining?.id
```
itsaky-adfa: NITPICK: deleting the last session leaves `_currentSessionId` null, and the new test pins that state as correct.

`remaining?.id` is null when nothing survives, after which `replaceCurrentSessionMessages` and `syncMessageToSession` return early on their `?: return`, so every later message renders and is never stored. The line is pre-existing and unreachable while `deleteSession` has no caller, so nothing ships wrong today -- but `givenTheLastSessionDeleted_whenNoneRemains_thenHistoryIsEmpty` now asserts the null-session outcome, so whoever wires the delete gesture will read the test as saying this is intended.

Call `createNewSession()` when `remaining` is null, and have the test assert that instead.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
itsaky-adfa: Re-checked at `f566cb6`: `clearMessages()` is fixed, `deleteSession` is not and the PR body was not updated. Carrying that half forward at https://github.com/appdevforall/plugin-examples/pull/99#discussion_r4046281047 rather than reopening it here.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
itsaky-adfa: Re-checked at `f566cb6`: the localized-`actionFailedText` case is fixed and pinned by a test. The `?:` does not catch the empty string the native `respond` path stores, so this became a CRITICAL in the new round: https://github.com/appdevforall/plugin-examples/pull/99#discussion_r4046281000

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
itsaky-adfa: Re-checked at `f566cb6`: both KDocs say messages now. The PR body still says "the last 40 turns", so the doc half is still open at https://github.com/appdevforall/plugin-examples/pull/99#discussion_r4046281030

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
itsaky-adfa: Re-checked at `f566cb6`: the character budget is in, but the loop spending it can return an empty history when the newest message alone exceeds it. Confirmed with a probe test; carried forward at https://github.com/appdevforall/plugin-examples/pull/99#discussion_r4046281019

itsaky-adfa (changes requested):
Blocking on the two findings in https://github.com/appdevforall/plugin-examples/pull/99#pullrequestreview-5247243120, both confirmed with probe tests at `f566cb6`:

- **`ChatViewModel.kt:1714`** (CRITICAL) - `historyText ?: text` falls back only on null, and the native `respond` path stores `""`. Every restored Gemini answer is a blank ASSISTANT turn.
- **`ChatViewModel.kt:1716`** (IMPORTANT) - one message longer than `MAX_RESTORED_HISTORY_CHARS` breaks the loop before anything is kept, so the restore is empty.

The four MINOR comments and the NITPICK are not blocking. Worth folding a test for each of the two into `ChatViewModelHistoryRestoreTest` while you are there: neither is covered today.

Everything else from the last round re-checked clean.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Fixed in d491f05 — the read is now `historyText?.takeIf { it.isNotBlank() } ?: text`, so a native `respond` with no text part falls back to the bubble.
`givenABlankHistoryText_whenRestoring_thenTheRenderedAnswerIsUsed` seeds `historyText = ""` and pins it.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Fixed in d491f05 — the budget check moved ahead of `chars +=` and is guarded by `kept.isNotEmpty()`, so the newest eligible turn is kept whatever it costs.
`givenANewestTurnOverTheWholeBudget_whenRestoring_thenItIsStillRestored` uses an 8001-character newest message, the shape yours did not reach.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Took your line verbatim: `it.sender == Sender.USER || (it.durationMs ?: 0L) > 0L`, so the null-duration bubble goes with the zero-duration one.
The test is renamed `givenAnUnfinishedAgentTurn_…` and covers both; the seed helper now stamps a duration on finished AGENT turns, as production does.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Added `cancelActiveRun("switch session")` and `cancelActiveRun("delete session")` in d491f05.
Agreed the guard belongs in the ViewModel rather than resting on `ChatFragment` calling `stopProcessing` first.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa `deleteSession` now calls `createNewSession()` when nothing remains — it cancels the run itself, so the epoch bump is not lost.
The test is `givenTheLastSessionDeleted_whenNoneRemains_thenAFreshSessionOpens` and asserts the fresh empty session instead of the null one.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Rewrote the How paragraph around the filter as it now stands, both caps (with the 40 stated as messages) and the newest-message exception.
The Details paragraph now lists what the suite actually covers.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Dropped that last round, sorry — the body now says `deleteSession` has no UI calling it and is API prepared for the delete gesture in a follow-up, exercised by tests only.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1686,77 @@ class ChatViewModel(
...
+                it.text.isNotBlank()
+        }
+        // Newest-first, so both budgets are spent on the turns nearest the next message.
+        val kept = ArrayDeque<LlmInferenceService.ChatMessage>()
+        var chars = 0
+        for (message in eligible.asReversed()) {
+            // What the model wrote, not the bubble: a turn whose tool call failed renders as
+            // "the action failed" in the IDE's language, a sentence the model never produced.
+            // Blank when a native respond call carried no text part, where the bubble is the answer.
+            val text = message.historyText?.takeIf { it.isNotBlank() } ?: message.text
```
Daniel-ADFA: IMPORTANT: the blank fallback re-injects the localized placeholder this whole mechanism exists to keep out of the model's context.

`AgentReplyRenderer.render` (`AgentReplyRenderer.kt:57`) returns `actionFailedText` from its *first* branch whenever a `respond` call follows a failed tool, before it ever looks at the model's prose. On a natively-calling backend a `respond` with no text part leaves `written = ""`, so `:1367` stores `text = "The action failed"` with `historyText = ""`. Here `"".takeIf { it.isNotBlank() }` is null, so the restore falls through to `message.text` and hands the model `ASSISTANT: "The action failed"` in the IDE's language - a sentence it never wrote, and one live `_history` never held either, since `AgentLoop.kt:179` skips a blank `historyText` entirely. A failed tool plus a native `respond` is an ordinary combination, not a corner.

`historyText` needs to separate "no model text was recorded" from "the model's text was empty"; the bubble is a safe fallback only when it is the model's own answer (`respondMessageOf`), never when it is `actionFailedText` or `noResponseText`.

`givenABlankHistoryText_whenRestoring_thenTheRenderedAnswerIsUsed` seeds a benign bubble, which is what makes the fallback look right. Seed it with `"The action failed"` - what production actually stores here - and it contradicts `givenATurnRenderedForDisplay_whenRestoring_thenTheModelsOwnTextIsUsed` directly above it.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1686,77 @@ class ChatViewModel(
...
+            // The newest eligible turn is kept whatever it costs: breaking on it would restore
+            // nothing at all, which is the regression this rebuild exists to remove.
+            if (kept.isNotEmpty() && (kept.size >= MAX_RESTORED_HISTORY ||
```
Daniel-ADFA, line 1702: IMPORTANT: the newest-turn exception is unbounded, so one oversized turn defeats the character budget for the rest of the session.

`kept.isNotEmpty()` is false on the first pass, so the newest eligible message is added whatever its length, and nothing upstream bounds that length: `prompt_input_edittext` (`fragment_chat.xml:167`) carries no `android:maxLength`, and `sendMessage` (:1016) rejects only a blank string. Paste a 200 KB file body into the composer and send it; switch away and back, or restart the IDE, and `_history` starts at 200 KB against a budget documented as 8,000 characters. It stays there: `_history` is only ever rebuilt by a restore, and `:1087` prepends it whole to every later send, so a local backend gets an over-context prompt on every turn for the rest of the session. At the merge base `switchToSession` emptied `_history`, so this is new.

Truncate the newest turn to `MAX_RESTORED_HISTORY_CHARS` rather than exempting it: that still avoids the empty restore the exception was added for. `givenANewestTurnOverTheWholeBudget_whenRestoring_thenItIsStillRestored` asserts the 8,001-character message comes back whole, so it pins the current behaviour and has to move with the fix.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1686,77 @@ class ChatViewModel(
...
+     * @return the eligible messages that fit both budgets, oldest first; the newest one is
+     *   always kept, however long it is.
+     */
+    private fun rebuildHistoryFrom(
+        messages: List<ChatMessage>
+    ): List<LlmInferenceService.ChatMessage> {
+        val eligible = messages.filter {
+            // SYSTEM notices and TOOL output are the scaffolding this rebuild exists to leave out.
+            (it.sender == Sender.USER || it.sender == Sender.AGENT) &&
+                // Only a finished agent turn carries a duration: null is a bubble process death
```
Daniel-ADFA, line 1686: MINOR: this comment names the process-death bubble, and the history side is now handled, but the same bubble still restores to the screen animating forever.

The defect is at `ChatAdapter.kt:205`, outside this diff and predating this PR - it does not block the merge. `onToken` (:1300) persists the streaming bubble through `syncMessageToSession` with `status = SENT` and `durationMs == null`; `finalizeInProgressMessages` runs only on Stop, so a force-stop or an OOM kill mid-answer leaves that row in the blob. On the next launch `bindDefaultMessage` matches `sender == AGENT && durationMs == null` under `SENT` and calls `animateGeneratingDots` - on a transcript with no run behind it and no Stop button to press. Still true at the stack tip.

Worth a ticket rather than this PR: finalize unfinished agent turns as the transcript loads, which is the same predicate this filter already computes.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1658,15 +1686,77 @@ class ChatViewModel(
...
+                // cut mid sentence, zero the marker Stop leaves; neither was finished saying.
+                (it.sender == Sender.USER || (it.durationMs ?: 0L) > 0L) &&
```
Daniel-ADFA, line 1688: MINOR: zero is not only the marker Stop leaves, so this also drops a turn that deliberately finished.

`addAgentMessage` (:1547) builds a terminal agent turn with `status = COMPLETED, durationMs = 0L`, and its own KDoc says why: "terminal state, no streaming dots". This filter reads that same zero as "never finished saying", so such a turn would show in the transcript and never reach the model. Unreachable today: grepping every `src/main` Kotlin file at this head and at the stack tip `5e0de33f` finds `addAgentMessage` only at its own declaration, so nothing can produce one. The first caller to wire it up loses its turns silently, against a comment asserting the two cases are exhaustive.

Either delete `addAgentMessage`, or give the Stop marker a state of its own - `finalizeInProgressMessages` writes `COMPLETED` too, so status alone cannot tell a stopped turn from a finished one either.

Daniel-ADFA (commented):
Re-review at `a10638e9`. All seven findings from the previous round are fixed - verified by reading the code at head, not the replies - and the re-check is below. Two new IMPORTANTs, both consequences of fixes made for that round.

**Stack.** This is the lower of two layers: `main` <- **#99** <- #101 (`feat/ADFA-6007-chat-session-list-ui`). The stack merges atomically, so every claim below was verified against the tip `5e0de33f`, not against this PR's head alone. Three defects that are real at this head are already repaired by #101; they are listed at the end rather than raised as comments, because they cannot reach `main`.

**Verdict rule.** This repo has no written approve/request-changes rule - `CLAUDE.md` covers the build and the plugin conventions, and `.claude/skills/plugin-review/` is a plugin-submission rubric, not a PR rule - so the default applied: a confirmed IMPORTANT blocks.

## Findings

**IMPORTANT**
- `ChatViewModel.kt:1717` - the blank fallback re-injects the localized "The action failed" as the model's own turn
- `ChatViewModel.kt:1720` - the newest-turn exception is unbounded, so one oversized paste defeats the character budget for the whole session

**MINOR**
- `ChatViewModel.kt:1706` - `durationMs > 0` also drops a turn that deliberately finished with zero
- `ChatViewModel.kt:1704` - a process-death bubble still restores to the screen animating forever (predates this PR, does not block it)

## Re-check of the previous round, all at `a10638e9`

- **CRITICAL `:1714`** blank `historyText` on the native `respond` path - **fixed for the case it named**, and the fix is what the first IMPORTANT above is about: `isNotBlank()` cannot tell "the bubble is the model's answer" from "the bubble is a localized placeholder", and falls back in both.
- **IMPORTANT `:1716`** a single over-budget message restored an empty history - **fixed.** The budget check moved ahead of `chars +=` and is guarded by `kept.isNotEmpty()` at `:1720`, pinned by `givenANewestTurnOverTheWholeBudget_whenRestoring_thenItIsStillRestored`. That guard is what the second IMPORTANT above is about.
- **MINOR `:122`** stale PR description - **fixed.** The body now describes the `durationMs` filter, both caps with the 40 stated as messages, and the newest-message exception; the Details paragraph matches what the suite covers.
- **MINOR `:1675`** `switchToSession` and `deleteSession` lacked the cancel - **fixed.** `cancelActiveRun("switch session")` at `:1670`, `cancelActiveRun("delete session")` at `:1753`. See the stack note below for what that cancel still omits.
- **MINOR `:1704`** the marker covered Stop only, not process death - **fixed** for the history. The filter is `(it.durationMs ?: 0L) > 0L`, and `givenAnUnfinishedAgentTurn_whenRestoring_thenItIsExcludedFromHistory` seeds both the zero and the null bubble. The screen half of it is the last finding above.
- **MINOR `:1745`** `deleteSession` had no production caller and the body did not say so - **fixed.** Still no caller - a grep over every `src/main` Kotlin file at this head finds only the declaration, the `ai-agent-mcp` hits being an unrelated HTTP method - and the body now discloses it as API for a follow-up.
- **NITPICK `:1742`** deleting the last session left `_currentSessionId` null - **fixed.** `:1748` routes to `createNewSession()`, pinned by `givenTheLastSessionDeleted_whenNoneRemains_thenAFreshSessionOpens`.

## Checked and clear

- **The budget arithmetic**, apart from the exception above. `kept.size >= MAX_RESTORED_HISTORY` keeps exactly 40; `chars +=` after the check means a kept set that starts within budget stays within it; `break` over `continue` is right, since it keeps a contiguous newest suffix instead of a transcript with a hole in it; `asReversed()` + `addFirst` gives oldest-first.
- **"Every backend here flattens the array"** (`:1729`) holds. `GeminiBackend.buildContents` opens with the system prompt as a `user` turn and folds consecutive same-role turns, so a leading ASSISTANT merges rather than reaching the API first; `LocalLlmBackend` flattens to `<|im_start|>`; OpenAI accepts consecutive same-role turns.
- **`historyText` persistence.** Gson reflects over the field, so no serializer needed, and an older blob without it deserializes to null, which is what the KDoc claims.
- **A pre-`durationMs` transcript generation.** There isn't one: `durationMs` has been on `ChatMessage` and populated on completion since `2773dc3`, the commit that introduced the class. `git log --follow` lists `627fc16` for this file, but that commit only moved it from `ai-assistant/` to `ai-core/`.

## Real at this head, already repaired by #101 - not raised as comments

`cancelActiveRun` (:1636) is `stopProcessing` minus `finalizeInProgressMessages()` and minus the return to `AgentState.Idle`, which costs its new callers:

1. **A stranded animating bubble.** `switchToSession` (:1670) and `createNewSession` (:1650) cancel a streaming run without finalizing it, so the departing transcript keeps a `durationMs == null` bubble that animates its dots whenever the user returns to that session, across restarts. #101 routes both through `endRunBeforeSessionChange` -> `stopProcessing`, which finalizes before `_currentSessionId` moves.
2. **A Stop button with no run under it.** `switchToSession` and `deleteSession` never return to `AgentState.Idle`, so after switching mid-run the composer still shows Stop and the next tap stops a run that no longer exists instead of sending. #101's `stopProcessing` ends on Idle.
3. **A shared mutable list behind the StateFlow.** `deleteSession` (:1755) assigns `_messages.value = remaining.messages` without the `.toList()` snapshot `switchToSession:1673` takes two lines earlier. #101's `adoptSession` snapshots.

I also checked and dropped a fourth: `loadSessions` -> `switchToSession` -> `cancelActiveRun` looks like it would kill an in-flight run on every rebind, but `syncStorageToCurrentProject` (`ChatFragment.kt:241`) returns early when the project has not changed and calls `stopProcessing` itself when it has, so it is not reachable.

Verified by reading at `a10638e9` and `5e0de33f`; the unit suite was not run for this round.

Daniel-ADFA (changes requested):
## Verdict: request changes

Submitted on top of the findings review above (#pullrequestreview-5266637730). Both blocking
findings are confirmed, and both are consequences of the fixes made for the last round rather
than anything that was wrong before it - the filter, the budgets and the backend role handling
all check out, and the seven findings from that round are genuinely fixed.

- **`ChatViewModel.kt:1717`** - `historyText?.takeIf { it.isNotBlank() } ?: text` cannot tell "the
  bubble is the model's own answer" from "the bubble is a localized placeholder", and falls back
  in both. A `respond` after a failed tool renders as `actionFailedText`
  (`AgentReplyRenderer.kt:57`, the first branch), so a restore hands the model
  `ASSISTANT: "The action failed"` in the IDE's language - a sentence it never wrote, and one
  live `_history` never held, since `AgentLoop.kt:179` skips a blank `historyText` outright.
  This is the exact injection the `historyText` field was added to prevent.

- **`ChatViewModel.kt:1720`** - the newest-turn exception is unbounded, and nothing upstream
  bounds message length. One oversized paste restores over the 8,000-character budget and stays
  in `_history` for the rest of the session, since a restore is the only thing that rebuilds it.

Both have a test pointing the wrong way: `givenABlankHistoryText_whenRestoring_thenTheRenderedAnswerIsUsed`
seeds a benign bubble where production stores the localized string, and
`givenANewestTurnOverTheWholeBudget_whenRestoring_thenItIsStillRestored` asserts the oversized
message comes back whole. Both need to move with the fixes.

The two MINORs do not block. `:1706` is latent (`addAgentMessage` has no caller), and `:1704`
predates this PR - worth a ticket, not a change here.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @Daniel-ADFA Fixed in 3949182d — a non-null blank `historyText` now drops the turn in the filter, the same thing `AgentLoop:179` does live, so `actionFailedText` never reaches the model.
`null` still means the bubble is the model's own text, which keeps that fallback only where it is safe.
The test now seeds `"The action failed"` with `historyText = ""` — your shape — and asserts the turn is excluded.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @Daniel-ADFA Fixed in 3949182d — the newest turn is truncated to `MAX_RESTORED_HISTORY_CHARS` instead of exempted whole, so an oversized paste cannot outlive the restore.
It keeps the tail, the part nearest the next message; say the word if the head of a pasted trace is the more useful half.
`givenANewestTurnOverTheWholeBudget_…_thenItIsTruncatedToTheBudget` moved with it and asserts `content.length == 8_000`.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @Daniel-ADFA Agreed, out of scope here — `ChatAdapter.kt:205` predates this PR and the fix belongs with the transcript load, not the rebuild.
Untouched in 3949182d; I will raise a ticket for finalizing unfinished agent turns as the transcript loads, reusing the predicate this filter already computes.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @Daniel-ADFA Right, the two zeroes are indistinguishable. `addAgentMessage` has had no caller since #65, so 3949182d deletes it rather than complicating the predicate for a shape nothing can produce.
That removes the trap for whoever wires the first caller; happy to bring it back with a real duration stamped if you would rather keep the helper.

Daniel-ADFA (approved):

itsaky-adfa (approved):

PR #100, opened 2026-09-16
ADFA-6041 | Implement agent cycle detection and fix tool approval integrity

jatezzz:
## Description

This PR addresses the issue of the agent getting stuck in unproductive loops and ensures strict approval gating for tools.

* **What**: Added cycle detection to stop the agent after 3 turns without new tool calls (`StopReason.CYCLING`). Refactored tool approvals to rely solely on the handler's `requiresApproval` property, completely removing the `AUTO_APPROVED_TOOLS` name-based bypass.
* **How**: Introduced `ToolCallProgressGuard` to monitor for cyclical, non-progressive tool signatures. Set `requiresApproval = true` for `GenerateFromTemplateHandler` and `GradleSyncHandler`. Added an `AgentRunReporter` to decouple tracing and UI notices from the `ChatViewModel`.
* **Why**: To prevent the agent from wasting step budgets and backend API costs on infinite read-loops, and to guarantee that any tool modifying the project or triggering a build strictly requires explicit user consent before executing.

## Details

* Added `onNoProgressCycle` event and a corresponding UI message in `strings.xml`.
* Test suites were updated (`AgentLoopTest.kt`, `ToolApprovalManagerTest.kt`, `ToolCallProgressGuardTest.kt`, and `BuiltInHandlerApprovalTest.kt`) to validate the new cycling guard and ensure no write-tools can bypass approval.
* Logic handling the intersection of `onRepeatAfterSuccess` and cycling loops has been resolved (cycling checks evaluate first).

https://github.com/user-attachments/assets/a9dae9d6-7c32-4f0c-9985-9eb7f8bc0683

## Ticket

[ADFA-6041](https://appdevforall.atlassian.net/browse/ADFA-6041)

## Observation

The `gradle_sync` tool and `generate_from_template` tool will now explicitly prompt the user for approval. Extracted run narration into `AgentRunReporter` to keep trace vocabulary and UI wording isolated from the view model.

[ADFA-6041]: https://appdevforall.atlassian.net/browse/ADFA-6041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuard.kt
```diff
@@ -0,0 +1,95 @@
...
+        previousBatchSucceeded = results.isNotEmpty() && results.all { it.success }
+    }
+
+    private fun verdictFor(signature: String): Verdict {
+        if (seenSignatures.add(signature)) {
+            turnsWithoutNewSignature = 0
+        } else {
+            turnsWithoutNewSignature++
+            // Ahead of the repeat checks: a run that has circled this long has not finished.
+            if (turnsWithoutNewSignature >= maxTurnsWithoutProgress) return Verdict.CYCLING
```
itsaky-adfa, line 95: IMPORTANT: a run that verifies its own edits trips this and is reported to the user as having made no progress. `turnsWithoutNewSignature` counts any turn whose batch signature appeared earlier in the run, with no notion of whether work happened in between, so three consecutive re-reads abort a run that succeeded.

I ran this against `fc116ac` as a JUnit case on the real `AgentLoop`:

```
T1 search_project(foo)  new     T5 edit_file(B)   new
T2 read_file(A)         new     T6 read_file(A)   seen -> stale 1
T3 edit_file(A)         new     T7 read_file(B)   seen -> stale 2
T4 read_file(B)         new     T8 search_project(foo) seen -> stale 3 -> CYCLING
```

Result: `reason=CYCLING turns=8 batchesRun=7`. Both edits landed at T3 and T5, T8's tool never runs, and the user is shown `agent_no_progress`: "the model kept redoing actions it had already taken instead of making progress". Editing three files then re-reading all three trips it with no search step at all. `DEFAULT_MAX_ITERATIONS`'s own KDoc records that "at 8 a two-line rename ran out mid-way" - this now aborts a rename at turn 8 of 16 and calls it a failure.

Reset `turnsWithoutNewSignature` on any successful mutating batch, so novelty is only asked about a run that is not otherwise changing anything.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/AgentTools.kt
```diff
@@ -55,8 +55,6 @@ class AgentTools private constructor(
         ): AgentTools {
             val reserved = builtInHandlers.mapTo(mutableSetOf()) { it.toolName }
             reserved += terminalTool
-            // The approval gate exempts a name, not a handler, so those names are reserved too.
-            reserved += ToolApprovalManager.AUTO_APPROVED_TOOLS
 
             val handlers = builtInHandlers + store.handlers(reserved)
```
itsaky-adfa, line 59: MINOR: the set passed here shrank, but `ToolSourceStore.handlers`'s own KDoc still describes the old one. `ToolSourceStore.kt:112-113`, outside this diff, documents `reservedNames` as "the built-in tools, the terminal tool, and every auto-approved name, which the approval gate exempts whether a handler carries it or not".

Nothing exempts a name any more and `AUTO_APPROVED_TOOLS` is gone, so that sentence is the last surviving description of the mechanism this PR exists to delete, sitting on the function whose job is to enforce reservations. The README paragraph saying the same thing was updated; this one was missed, so the next person reasoning about why a name is reserved concludes the name-based exemption is still live.

Trim it to "the built-in tools and the terminal tool", matching what `build` passes.

ai-core/src/main/assets/docs/index.html
```diff
@@ -100,8 +100,9 @@ <h2>What the agent can do</h2>
       <b>Correct</b> to tell the agent what to change and let it try again.
       When the file is open in the editor, the change is applied to the buffer,
       so <b>Ctrl+Z undoes it</b> and your unsaved work is preserved.</li>
-    <li>Trigger a Gradle sync and read build output.</li>
-    <li>Add dependencies and generate code from templates.</li>
+    <li>Trigger a Gradle sync (asks for approval — a sync starts a real build)
+      and read build output.</li>
+    <li>Add dependencies and generate code from templates, both after asking.</li>
```
itsaky-adfa, line 105: NITPICK: this page and `ai-core.html` both learned that a sync and a template generation now ask, but the in-app tooltips did not. `AiCorePlugin.kt:282-283` still tells the user "Anything that writes to the project asks for your approval first", and `:244-245` says file-editing tools "ask for approval before writing".

Neither sentence is false, but a user who long-presses for help learns the rule is about writes, then gets a dialog for `gradle_sync`, which writes nothing they can see. Tier 1/2 tooltips are the surface most users actually reach; the offline page is the one they do not open.

Add "as does starting a Gradle sync or generating from a template" to the `:282` entry, mirroring the wording already landed in `ai-core.html`.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuard.kt
```diff
@@ -0,0 +1,95 @@
...
+    private val seenSignatures = mutableSetOf<String>()
+    private var previousSignature: String? = null
+    private var consecutiveRepeats = 0
+    private var turnsWithoutNewSignature = 0
+
+    // Null until a batch has run: "no tools yet" and "the tools failed" end a run differently.
+    private var previousBatchSucceeded: Boolean? = null
+
+    /** Turns running that have introduced no batch the run had not already issued. */
+    val staleTurns: Int get() = turnsWithoutNewSignature
```
itsaky-adfa, line 52: NITPICK: `staleTurns` can only ever hold one value at the one place that reads it. `verdictFor` increments by one and returns `CYCLING` the moment the counter reaches `maxTurnsWithoutProgress`, so `events.onNoProgressCycle(turn, progress.staleTurns)` always passes exactly that threshold.

Nothing misbehaves, but `AgentRunReporter` traces "no new tool action for 3 turns" on every cycling stop - it reads as a measurement and is a constant, so it is the one number in that line that cannot tell two runs apart.

Drop the parameter and let the reporter name the configured limit, or say in the KDoc that it equals the threshold by construction. (Moot if the counter changes.)

itsaky-adfa (commented):
Reviewed at `fc116ac` (high effort). The approval half of this PR is solid; one finding on the cycling half.

## Findings

**IMPORTANT**
- `ToolCallProgressGuard.kt:74` - a verify-after-edit sweep aborts a successful run as CYCLING

**MINOR**
- `AgentTools.kt:59` - `ToolSourceStore.handlers` KDoc still documents the deleted auto-approved reservation

**NITPICK** - 2 inline, not listed

## Verification

Claims were checked against the code at `fc116ac`, not taken from the diff alone.

- `./gradlew testDebugUnitTest` in `ai-core/`: **all pass**, including the four new suites.
- The IMPORTANT finding is reproduced by a JUnit case driving the real `AgentLoop`, not reasoned from the source: `reason=CYCLING turns=8 batchesRun=7`. That is the only reason it is graded IMPORTANT rather than PLAUSIBLE.

## Checked and cleared

- **No approval bypass survives.** `Executor.executeCall` always calls `ensureApproved`; `ContributedToolHandler` hard-codes `requiresApproval = true`; all 12 built-ins are classified and pinned by `BuiltInHandlerApprovalTest`, whose exhaustiveness assertion fails if a new handler is added unclassified. No `AUTO_APPROVED_TOOLS` reference remains anywhere.
- **Dropping `reserved += AUTO_APPROVED_TOOLS` is safe.** All seven names are `BuiltInToolHandlers` names, so `reserved` is unchanged whenever built-ins exist; with a null context there are no built-ins to shadow.
- **`AgentRunReporter` overrides 9 of 9 `Events` callbacks** - nothing was dropped moving them out of `ChatViewModel`, and both system-message strings are unchanged.
- `StopReason` has no exhaustive `when` anywhere, so `CYCLING` breaks no caller. One `AgentLoop` construction, one `Events` implementation.
- `previousSignature` updating only on `PROCEED` is correct, since every other verdict returns from the loop; `lastBatchFailed` preserves the tri-state `null`/`false` distinction; `inspect` is never handed an empty batch.
- `runNotices` is declared after its use site but read only from a suspend function, so there is no initialization-order hazard.
- The three doc surfaces that were updated (`README.md`, `ai-core.html`, `assets/docs/index.html`) are accurate against the new behaviour.

## Prior rounds

No prior inline findings to re-check. The one existing review is a `claude[bot]` configuration notice with no findings in it.

## Verdict basis

This repo has no written approve/request-changes rule (no `REVIEW.md` or `CONTRIBUTING.md`, and `CLAUDE.md` sets no threshold), so the default applied: a confirmed IMPORTANT blocks, MINOR and NITPICK do not.

itsaky-adfa (changes requested):
Blocking on one finding: `ToolCallProgressGuard` aborts a run that is verifying its own edits and reports it to the user as having made no progress. Reproduced on a JUnit case driving the real `AgentLoop` at `fc116ac`: three files edited then re-read gives `reason=CYCLING` with both edits already on disk and the last turn's tool dropped.

To clear this: reset `turnsWithoutNewSignature` on any successful mutating batch, so signature novelty is only asked about a run that is not otherwise changing anything, and add a case pinning that an edit-then-verify run still ends `COMPLETED`.

Everything else is fine. The approval half is well covered - no bypass survives and `BuiltInHandlerApprovalTest` pins the invariant for new handlers. The MINOR and the two NITPICKs are non-blocking; take or leave them.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuard.kt
jatezzz: @itsaky-adfa Fixed — a successful mutating batch now clears `seenSignatures` as well as resetting the counter; a bare reset is a no-op on your trace, since `edit_file(B)` at T5 was already novel.
An edit changes what a re-read returns, so T6–T8 are new actions and the run finishes; a read-only rotation still trips at 3, and a run that edits then spins gets 3 turns from the last edit.
`givenEditsFollowedByReReads_whenInspected_thenVerifyingItsOwnWorkIsNotCycling` covers your T1–T8.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/AgentTools.kt
jatezzz: @itsaky-adfa Trimmed to "the built-in tools and the terminal tool", matching what `build` passes.

ai-core/src/main/assets/docs/index.html
jatezzz: @itsaky-adfa Added to the `:282` entry: it now reads "…asks for your approval first, as does starting a Gradle sync or generating from a template", mirroring `ai-core.html`.
Left `:244` alone — it is scoped to file-editing tools, so it stays accurate.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuard.kt
jatezzz: @itsaky-adfa Renamed the parameter to `staleLimit`, with KDoc saying it equals the configured threshold by construction; the trace now reads "no new tool action for the configured 3 turns".
Kept the parameter rather than dropping it — `AgentRunReporter` has no other way to name a non-default limit.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1101,82 +1067,11 @@ class ChatViewModel(
...
-                            override suspend fun onRepeatedToolCalls(turns: Int) {
-                                AgentTrace.refusal("LOOP", "turns=$turns", "identical tool calls repeated")
-                                addSystemMessage(
-                                    str(R.string.agent_repeated_calls),
-                                    MessageStatus.SENT
-                                )
-                            }
-                        }
+                        // A handler that asks for approval is one that changes the project.
+                        isMutatingTool = { name ->
```
itsaky-adfa: IMPORTANT: every contributed tool counts as mutating here, so the cycling guard never fires for a run that loops over another plugin's tools.

`ContributedToolHandler.requiresApproval` is hard-coded `true` whatever the source declared (`ContributedToolHandler.kt:62`), so this returns true for every MCP/plugin tool, read-only ones included. Each successful turn then clears `seenSignatures` in `ToolCallProgressGuard.recordResults`, and nothing is ever stale.

Reproduced as a JUnit case on the real `AgentLoop` at `9028c8ba`: a model rotating `mcp_search_docs` / `mcp_get_page` / `mcp_list_repos` runs to `MAX_ITERATIONS` at 16 turns, while the identical rotation over `read_file` / `list_files` / `search_project` stops as `CYCLING`. The contributed surface is where a read-loop burns a paid backend, and the docs edited in this PR recommend installing AI Agent MCP.

`ContributedToolHandler.readOnly` already carries the provider's declaration - derive mutation from that, or from a new `ToolHandler.mutatesProject`, rather than from the approval flag, which is a security default and not a progress signal.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuard.kt
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Records how the batch just run turned out.
+     * @param results the batch's results, as the loop received them.
+     */
+    fun recordResults(results: List<ToolResult>) {
+        previousBatchSucceeded = results.isNotEmpty() && results.all { it.success }
+        // An edit makes earlier calls stale rather than repeatable: re-reading a file it changed
+        // is a new action, so a run that edits and then verifies is not judged as going in circles.
+        if (previousBatchSucceeded == true && currentBatchMutates) {
+            seenSignatures.clear()
```
itsaky-adfa: IMPORTANT: clearing the whole seen-set makes a run of nothing but mutating calls invisible to both guards.

`clear()` drops every signature, not just the ones the edit invalidated. A model flip-flopping `edit_file(A.kt, X)` / `edit_file(A.kt, Y)` succeeds every turn, so the set is emptied every turn and no signature is ever seen again (no CYCLING); consecutive signatures always differ (no REPEATED).

On the real `AgentLoop` at `9028c8ba` that run reaches `MAX_ITERATIONS` having executed all 16 edits, rewriting one file back and forth with an approval prompt each time. The same sequence with `isMutatingTool = { false }` stops as `CYCLING` on turn 5 - the escape hatch added this round is what hides it. An `edit_file`/`read_file` ping-pong on one path behaves the same way.

Clearing only the signatures naming a path the batch touched, and keeping the mutating signature itself in the set, would close both.

itsaky-adfa (commented):
Second round. All four findings from the first round are fixed - verified by reading the code at `9028c8ba`, not from the replies:

- `ToolCallProgressGuard.kt:87` (IMPORTANT, a run verifying its own edits aborted as cycling) - **fixed**. A successful mutating batch now clears `seenSignatures` as well as resetting the counter, so the T1-T8 trace I gave finishes; `givenEditsFollowedByReReads...` pins it. See the second IMPORTANT below, though: that clear is wider than it needs to be.
- `AgentTools.kt:59` (MINOR, stale `reservedNames` KDoc) - **fixed**. `ToolSourceStore.kt:112` now reads "the built-in tools and the terminal tool", matching what `build` passes. Every removed `AUTO_APPROVED_TOOLS` name was already a built-in tool name, so the reserved set is unchanged in content.
- `assets/docs/index.html:105` (NITPICK, tooltip out of step with the docs) - **fixed** at `AiCorePlugin.kt:282`. Leaving `:244` alone is right; it is scoped to file-editing tools.
- `ToolCallProgressGuard.kt:47` (NITPICK, `staleTurns` argument) - **fixed**, renamed `staleLimit` with the invariant stated in the KDoc.

Two new findings. Both are reproduced as JUnit cases against the real `AgentLoop` at `9028c8ba`, and both are the same root cause seen from two sides: what the guard treats as "progress was made".

**CRITICAL** - none

**IMPORTANT**
- `ChatViewModel.kt:1071` - every contributed tool counts as mutating, so cycling detection never fires for an MCP read-loop
- `ToolCallProgressGuard.kt:76` - the total seen-set clear makes a run of nothing but edits invisible to both guards

**MINOR** - none

**NITPICK** - none

The approval half of the change is sound and I could not fault it: `ensureApproved` at `Executor.kt:224` is the only gate and sits ahead of every dispatch, every built-in that acts on the project declares `requiresApproval`, and `BuiltInHandlerApprovalTest` pins both halves of that rule against the same catalogue the chat registers rather than a copy. `ai-core.html`'s "Only reads run unprompted" is accurate as written. `AgentRunReporter` preserves all eight previous `Events` overrides.

`./gradlew testDebugUnitTest` at `9028c8ba`: **342 tests, 0 failures**. Not device-verified - per this repo's `CLAUDE.md` that means the change is not verified. The new approval dialogs for `gradle_sync` and `generate_from_template`, and the no-progress notice, still want a device pass.

This repo has no written approve/request-changes rule, so the default applied: two confirmed IMPORTANT findings.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @itsaky-adfa Fixed — mutation now comes from a new `ToolHandler.mutatesProject`, which `ContributedToolHandler` overrides with `!readOnly`, so the forced approval flag no longer stands in for a progress signal.
It defaults to `requiresApproval`, which is exactly right for the built-ins `BuiltInHandlerApprovalTest` already pins, so no handler had to change.
A read-only MCP rotation now goes stale the same way the `read_file`/`list_files`/`search_project` one does.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuard.kt
jatezzz: @itsaky-adfa Fixed — the guard now takes the paths a call changes (the handler's `pathArgs`) and evicts only seen signatures naming one of them, never a signature that itself changed a file.
It also evicts only on a change the run had not already made, so the `edit_file(A.kt,X)`/`edit_file(A.kt,Y)` flip-flop and the `edit_file`/`read_file` ping-pong both stop as CYCLING on turn 5.
`givenOneFileRewrittenBackAndForth_whenInspected_thenItStopsAsCycling` and `givenAnEditReissuedBetweenReReads_whenInspected_thenItStopsAsCycling` cover them; your earlier T1–T8 trace still finishes.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
```diff
@@ -1101,82 +1067,13 @@ class ChatViewModel(
...
-                                    str(R.string.agent_repeated_calls),
-                                    MessageStatus.SENT
-                                )
-                            }
-                        }
+                        // The paths a changing call rewrites, so a re-read of one counts as new.
+                        mutatedPathsOf = { call ->
+                            val handler = tools.router.getHandler(call.name)
+                            if (handler?.mutatesProject != true) emptySet()
+                            else handler.pathArgs.mapNotNull { call.args[it]?.toString() }.toSet()
```
Daniel-ADFA: IMPORTANT: this reads the model's raw argument keys, so a write whose path arrives under any other spelling contributes no path and the guard treats the turn as a read — which brings back the abort this round exists to fix.

Two spellings a real run produces:
- `edit_file`/`create_file`/`update_file` emitted with `path`. `Executor.executeCall` remaps `path` → `file_path` into a *copy* (`Executor.kt:157-161`), so the edit lands, but `ToolCall.args` still has no `file_path` and `pathArgs` names only that. The remap and `EditFileHandler.argAliases` both exist because small models do emit the near-miss key.
- `add_dependency` without `build_file`: the schema marks it optional and the handler defaults it to `app/build.gradle.kts` (`AddDependencyHandler.kt:40-41`), so the ordinary call carries no path at all. Every contributed write tool is the same — `ContributedToolHandler` declares `mutatesProject` but no `pathArgs`.

Run your own T1–T8 trace with `add_dependency` in place of `edit_file`: T3 and T5 mutate but evict nothing, T6/T7/T8 are all seen, and it stops as `CYCLING` at turn 8 of 16 with both dependencies added and `agent_no_progress` on screen.

Derive the paths from the same normalization the executor applies, and give a handler a way to name the path it defaults to.

ai-core/src/test/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuardTest.kt
```diff
@@ -0,0 +1,173 @@
...
+ * which guard wins when two could fire, and when a counter goes back to zero.
+ */
+class ToolCallProgressGuardTest {
+
+    private fun guard(repeats: Int = 2, stale: Int = 3, mutating: Set<String> = setOf("edit_file")) =
+        ToolCallProgressGuard(
+            maxConsecutiveRepeats = repeats,
+            maxTurnsWithoutProgress = stale,
+            mutatedPathsOf = { call ->
+                if (call.name in mutating) setOfNotNull(call.args["path"]?.toString()) else emptySet()
```
Daniel-ADFA: MINOR: this stub keys on `path`, which no production handler declares. Every `pathArgs` in the plugin is `file_path`, `build_file`, `directory` or `project_dir`, so the four tests that pin the edit-then-re-read reset exercise a key the real `mutatedPathsOf` never reads.

Nothing here fails, and that is the problem: the production lambda at `ChatViewModel.kt:1071-1075` has no test of its own, so the two empty-path cases in the thread above pass the whole suite. A reviewer reading `givenEditsFollowedByReReads_...` reasonably concludes the wiring is covered.

Key the stub on `file_path` and add one case over `BuiltInToolHandlers.create(...)` handlers, so the lambda is tested against the argument names the handlers actually declare.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuard.kt
```diff
@@ -0,0 +1,116 @@
...
+     * Records how the batch just run turned out.
+     * @param results the batch's results, as the loop received them.
+     */
+    fun recordResults(results: List<ToolResult>) {
+        previousBatchSucceeded = results.isNotEmpty() && results.all { it.success }
+        // A change the run had not made before makes an earlier read of the paths it touched a new
+        // action again, so a run that edits and then verifies is not judged as going in circles.
+        if (previousBatchSucceeded == true && currentBatchIsNew && currentBatchPaths.isNotEmpty()) {
+            seenSignatures.removeAll { seen ->
+                seen !in changingSignatures && currentBatchPaths.any { seen.contains(it) }
```
Daniel-ADFA: NITPICK: `seen.contains(it)` tests the path against the whole signature, argument *values* included, not against the path arguments.

So an `edit_file` whose `new_string` quotes `Foo.kt` evicts every seen signature mentioning `Foo.kt`, and a short path is a substring of longer ones (`A.kt` matches `bundle/A.kt2`). It errs toward too much novelty, so the cost is a run that circles for longer rather than one aborted early — but it makes the rule hard to reason about from the code.

Keep the per-call path set beside its signature and compare sets, rather than substring-matching the rendered string.

Daniel-ADFA (commented):
This is a re-review of the two prior rounds plus a fresh pass. Verified against head `5c3927f`; the repo has no written approve/request-changes rule (`plugin-examples/CLAUDE.md` covers build, verification and plugin shape only), so the default severity table applied.

## Severity index

**IMPORTANT**
- `ChatViewModel.kt:1074` - a write whose path arrives under another spelling contributes no mutated path, so the cycling guard treats it as a read

**MINOR**
- `ToolCallProgressGuardTest.kt:20` - the reset tests stub an argument key no production handler declares, so the real wiring is untested

**NITPICK** - 1 inline

## Re-check of the earlier rounds

All six prior findings are genuinely fixed; each checked by reading the code at head, not the replies.

- **#4039259750** (a run verifying its own edits trips the guard) - fixed. `recordResults` now evicts seen signatures naming a path the successful batch changed, and `givenEditsFollowedByReReads_...` re-runs your T1-T8 trace to `PROCEED` at T8. Traced by hand: T3 and T5 evict the reads of A and B, so T6-T8 are novel again.
- **#4039259758** (`ToolSourceStore.handlers` KDoc) - fixed. `ToolSourceStore.kt:113` now reads "the built-in tools and the terminal tool", matching what `AgentTools.build` passes.
- **#4039259763** (tooltips not updated) - fixed. `AiCorePlugin.kt:283-284` now carries "as does starting a Gradle sync or generating from a template". `:244` was correctly left alone.
- **#4039259772** (`staleTurns` is a constant) - fixed as documented rather than removed: renamed `staleLimit`, KDoc says it equals the threshold by construction, and `AgentRunReporter.kt:94` reads "for the configured N turns".
- **#4046272809** (every contributed tool counted as mutating) - fixed. `ToolHandler.mutatesProject` defaults to `requiresApproval` and `ContributedToolHandler` overrides it with `!readOnly`, so the forced approval flag is no longer a progress signal.
- **#4046272815** (clearing the whole seen-set) - fixed. Eviction is now scoped to the batch's paths, and `changingSignatures` keeps a mutating signature out of its own eviction. Both flip-flop traces stop as `CYCLING` on turn 5; `givenOneFileRewrittenBackAndForth_...` and `givenAnEditReissuedBetweenReReads_...` pin them.

The one finding this round is the unfinished half of that last fix: the eviction is right, but the path set feeding it is empty for more calls than it looks.

## The approval half

Checked separately and it is sound. Every one of the seven `AUTO_APPROVED_TOOLS` names is a built-in, so dropping the set from `AgentTools.build` loses no reservation - `reserved` is unchanged in practice, and `givenAToolForEveryReservedName_...` now pins that directly. `ToolRouter.dispatch` is reached only from `Executor.executeCall`, after `ensureApproved`, so there is no second path to a handler. `BuiltInHandlerApprovalTest` pins both directions and fails when a new built-in is added unclassified.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolHandler.kt
```diff
@@ -30,6 +30,14 @@ interface ToolHandler {
     val requiresApproval: Boolean
         get() = false
 
+    /**
+     * Whether this tool changes the project, which is what the progress guard reads. Defaults to
+     * [requiresApproval], true of exactly the built-ins that act on the project; a source whose
+     * approval flag is forced true overrides this instead.
+     */
+    val mutatesProject: Boolean
```
Daniel-ADFA, line 38: MINOR: defaulting `mutatesProject` to `requiresApproval` labels three build tools as changing the project when none of them writes a project file — `run_app`, `gradle_sync`, and `generate_from_template`, whose own KDoc says `execute` writes nothing today.

No current caller is affected: all three declare no `pathArgs`, so `mutatedPathsOf` returns an empty set and the mislabel is inert. It stops being inert the moment one of them gains a `pathArgs` entry, or a future built-in needs approval but only reads — it would then evict seen signatures on every successful turn and quietly disable the cycling guard for the rest of the run.

`ContributedToolHandler` already had to override this default for exactly this reason. That is the signal the default is wrong rather than a coincidence worth keeping: make `mutatesProject` abstract, or default it to false and declare it on the four handlers that write.

Daniel-ADFA (changes requested):
Blocking on the one IMPORTANT in the review above, plus one MINOR inline here.

**To unblock:** `ChatViewModel.kt:1074` builds the guard's path set from the model's raw argument keys, so a write whose path arrives under another spelling contributes nothing and the run is judged as circling. Deriving it from the same normalization the executor applies closes it.

**One addition to the substring nitpick** (`ToolCallProgressGuard.kt:83`): the direction that costs more is the *under*-match, not the over-match I named there. `read_file(file_path=app/src/Main.kt)` followed by `edit_file(file_path=./app/src/Main.kt)` is one file through `PathGuard`, but `./app/src/Main.kt` is not a substring of the read's signature, so the read is never evicted and the verify-after-edit re-read still counts stale. Same root cause as the blocking finding - a raw model-supplied string is being used where a resolved path is meant - and the same fix covers both: resolve the path, keep it beside its signature, and compare sets.

**What I checked and cleared**, so it is not re-litigated next round: every one of the seven former `AUTO_APPROVED_TOOLS` names is a built-in, so dropping the set from `AgentTools.build` loses no reservation; `ToolRouter.dispatch` is reachable only from `Executor.executeCall` after `ensureApproved`; all five former auto-approved read tools declare `requiresApproval = false`; and `ai-core`'s unit tests compile and pass at this head (`compileDebugUnitTestKotlin`, `testDebugUnitTest`). All six findings from the previous two rounds are genuinely fixed - detail in the review above.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/viewmodel/ChatViewModel.kt
jatezzz: @Daniel-ADFA Fixed — paths now come from `ToolHandler.pathsIn`, which applies the executor's own normalization (extracted to `normalizeToolArgs`, so there is one copy) and then a new `ToolHandler.pathDefaults`.
`AddDependencyHandler` declares `build_file` → `app/build.gradle.kts`, so the ordinary call names a path and a re-read of the build file is a new action again.
Your T1–T8 with `add_dependency` still ends at T8, but only because its re-reads are of `A.kt`/`B.kt`, which the dependency writes never touched — re-reading what the write changed is what the eviction now covers.

ai-core/src/test/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuardTest.kt
jatezzz: @Daniel-ADFA Fixed — the stub is keyed on `file_path`, and a `realGuard()` helper wires the guard over `BuiltInToolHandlers.create(...)` the same way `ChatViewModel` does.
Two cases over those handlers cover the holes from the thread above: `edit_file` spelled `path`, and `add_dependency` with no `build_file`.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolCallProgressGuard.kt
jatezzz: @Daniel-ADFA Fixed — `seenSignatures` is now signature → the paths that batch named, and eviction compares sets instead of substring-matching the rendered signature.
That needed paths for reads too, so `mutatedPathsOf` split into `pathsOf` + `changesPaths`; a `new_string` quoting `Foo.kt` no longer evicts, and `A.kt` no longer matches `bundle/A.kt2`.
A change whose handler names no path (contributed writes, `gradle_sync`) still invalidates every non-changing signature — there is no way to tell which reads it stood for.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/tool/ToolHandler.kt
jatezzz: @Daniel-ADFA Fixed — `mutatesProject` now defaults to false and is declared on the four handlers that write (`create_file`, `update_file`, `edit_file`, `add_dependency`).
`run_app`, `gradle_sync` and `generate_from_template` no longer read as changes, and `ContributedToolHandler` keeps its `!readOnly` override.
`BuiltInHandlerApprovalTest` pins that set, so a new built-in cannot quietly re-acquire the old default.

Daniel-ADFA (approved):

PR #101, opened 2026-09-17
ADFA-6007 | Implement chat sidebar for per-project session management

jatezzz:
## Description

Introduces a sliding sidebar to replace the previous chat overflow menu, allowing users to manage their per-project chat sessions directly within the UI. Users can now seamlessly browse, switch to, rename, and delete their previous conversations. Crucially, the actions for starting a "New chat" (which preserves the previous thread) and "Clear this chat" (which empties it) are now distinct and explicitly separated to prevent accidental history loss.

## Details

* **Sidebar Integration:** Replaced the `ChatSessionsDialogFragment` with `ChatSidebarController` to manage the overlay panel, its lifecycle, and slide animations.
* **Layout Structure:** Updated `fragment_chat.xml` to include `view_chat_sidebar.xml` as an overlaid element rather than taking layout weight.
* **Session Paging:** Implemented `SessionPaging` logic to progressively load rows as the user scrolls, avoiding performance drops on projects with extensive histories.
* **Bulk Actions:** Added long-press multi-selection mode to the session list, allowing users to select and delete several chats at once.
* **Tooltips & Navigation:** Updated `AiCorePlugin` with new tooltips for sidebar actions and linked the sidebar button to the new view.

https://github.com/user-attachments/assets/a5e90806-3de7-4ea6-bd3d-9f8d437e7e76

https://github.com/user-attachments/assets/9f3fae89-68f6-4cd6-b149-e8342f95e883

## Ticket

[ADFA-6007](https://appdevforall.atlassian.net/browse/ADFA-6007)

## Observation

The sidebar is overlaid directly on the chat column rather than docked beside it; this prevents the layout pass from reducing the `RecyclerView` height to zero when the interface is displayed on shorter screens or in landscape mode. Nested scrolling was also explicitly disabled on the sidebar's list to prevent touch events from dragging the IDE's bottom sheet closed.

Must be merged first:
* https://github.com/appdevforall/plugin-examples/pull/99 

[ADFA-6007]: https://appdevforall.atlassian.net/browse/ADFA-6007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/ChatFragment.kt
```diff
@@ -136,6 +151,8 @@ class ChatFragment : Fragment(), ApprovalDialogFragment.Host {
         }
         composer?.detach()
         composer = null
+        sidebar?.detach()
```
hal-eisen-adfa, line 154: **F01 (medium): Back button becomes dead after the view is destroyed with the sidebar open.**

`sidebarBackCallback.isEnabled` is flipped only by `onOpenChanged`. `detach()` here does not fire that callback, so the flag stays `true`.

Destroy the view with the panel open (the bottom-sheet tab switch the class docs call out), then come back: `setupSidebar()` re-registers the same callback, still enabled, against a fresh *closed* controller. `handleOnBackPressed` calls `sidebar?.close()`, which returns `false` and does nothing — but the dispatcher treats the press as consumed. Back stays dead until the user opens and closes the panel once.

`handleOnBackPressed` also discards `close()`'s documented `true`/`false` return, so it cannot forward an unhandled press.

Fix: set `sidebarBackCallback.isEnabled = false` in `onDestroyView`, or have `detach()` invoke `onOpenChanged(false)`.

hal-eisen-adfa (commented):
Three findings from a review of this PR's own commit.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/ChatSidebarController.kt
```diff
@@ -0,0 +1,529 @@
...
+        childDialog = null
+        _binding = null
+    }
+
+    /**
+     * Re-measures the panel after a rotation. EditorActivity handles orientation itself, so nothing
+     * here is recreated and the width worked out for the old orientation would otherwise stand —
+     * either overhanging the screen or leaving most of it bare.
+     */
+    fun onConfigurationChanged() {
```
hal-eisen-adfa, line 176: **F02 (medium): the panel is re-measured from the pre-rotation width.**

`onConfigurationChanged()` recomputes the width synchronously through `panelWidth()`, which reads `_binding?.root?.width`. `ChatFragment.onConfigurationChanged` calls this *before* its own `binding.root.post { ... }` block — and that block exists precisely because the window has not published the rotated geometry yet.

So the re-layout uses the old width: portrait to landscape leaves the panel at the previous cap, and landscape to portrait can leave it wider than `available - chat_sidebar_min_scrim`, removing the scrim strip that is the documented close target.

Fix: post the re-measure (or use `doOnLayout`) so it runs against the rotated geometry.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/ChatSidebarController.kt
```diff
@@ -0,0 +1,529 @@
...
+
+    /**
+     * Republishes the list on every change to the sessions, which one is active, what is ticked or
+     * how much has been paged in — so a row retitles itself as a reply streams in, the tick follows
+     * a switch made from anywhere, and a page appended by scrolling goes through the same path as
+     * everything else.
+     */
+    private fun observeSessions() {
+        val untitled = _binding?.root?.context?.getString(R.string.session_untitled) ?: return
+        scope.launch {
```
hal-eisen-adfa, line 328: **F03 (medium): the session collector runs per streamed token while the panel is hidden.**

This `combine(...).collect` is launched on `viewLifecycleOwner.lifecycleScope` with no gate on `isOpen`. `ChatViewModel.replaceCurrentSessionMessages` rewrites `_sessions.value` once per streamed token, so every token re-sorts the whole session list (`newestFirst()`), rebuilds up to `limit` rows, starts an `AsyncListDiffer` diff, and touches views in `updateChrome` — all while the sidebar is `GONE`.

That is the same per-token cost `SessionPaging` was added to avoid, just moved to a different place.

Fix: collect only while the panel is open (`flowWithLifecycle` plus an `isOpen` gate, or start/stop the collection in `open()`/`close()`).

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/ChatFragment.kt
jatezzz: @hal-eisen-adfa Fixed in bcfa6573. `detach()` now fires `onOpenChanged(false)` when the panel was still open, so the fragment disarms `sidebarBackCallback` along with the views.
Doing it in the controller covers every caller and keeps the flag derived from one source, so the callback can no longer disagree with a fresh controller.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/ChatSidebarController.kt
jatezzz: @hal-eisen-adfa Fixed in bcfa6573. `sidebar?.onConfigurationChanged()` moved into the `binding.root.post { }` block that was already there for the cutout, so `panelWidth()` reads the rotated width.
The KDoc now states that obligation on the callee.

ai-core/src/main/kotlin/com/itsaky/androidide/plugins/aicore/fragments/ChatSidebarController.kt
jatezzz: @hal-eisen-adfa Fixed in bcfa6573. `isOpen` is backed by a `MutableStateFlow` and the collector is now `openState.flatMapLatest { if (!open) emptyFlow() else combine(...) }`, so `viewModel.sessions` is not collected at all while the panel is hidden.
`open()` also sets the flag after resetting `visibleCount`/`selection`, so the restarted collector never builds the previous browse pages first.

hal-eisen-adfa (approved):

